PR #1, opened 2020-05-13
Base structure for certificate provider

seocam:
The actual structure for the provider is in the commit `Base certificate management with certmonger`. The other commits have templates, lint configuration, .gitignore updates and examples.


seocam:
The idea of this structure is to have one a class to manage each certificate request that class is called 'CertificateRequestModule'.

Inside this class it verifies which provider should be used and based on that instantiate the provider class and then call a `run` method.

The provider class needs to implement an abstract class called `CertificateRequestCertmongerProvider`. The only two methods that needs to be implemented are `request_certificate` and `get_existing_csr_pem_data`. 

`get_existing_csr_pem_data` is should return the content of the CSR in text-pem format.
`request_certificate` is the command that actually issues a new certificate. It shouldn't need to worry about idempotency at all because that should be handled by the base provider (or mostly).

To help with the idempotency implementation I've add another class called "CertificateProxy"; that class can be instantiated using either the module params or the actual PEM from CSR and certificate. With that we can have a single object that represents a certificate and can be compared for idempotency purposes.

handlers/main.yml
```diff
@@ -0,0 +1,10 @@
+# SPDX-License-Identifier: MIT
+---
+# Put handlers for tasks here.
+
+# Example:
+- name: template handler to restart services
+  service:
+    name: "{{ item }}"
+    state: restarted
+  loop: "{{ __template_services }}"
```
richm, line 10: remove the `# Example:` line
change the `name` to be `certificate handler` instead of `template handler`
use `__certificate_services` instead of `__template_services`

meta/main.yml
```diff
@@ -0,0 +1,72 @@
+# SPDX-License-Identifier: MIT
+---
+galaxy_info:
+  # Replace with role's author name:
+  author: John Doe <jdoe@corp.com>
```
richm, line 5: remove the comments and fill in the fields with the actual data

tasks/main.yml
```diff
@@ -0,0 +1,25 @@
+---
+
+- name: Find required providers
+  set_fact:
+    providers: |
```
richm, line 5: This should be `__certificate_providers` - unless you intend this to be a variable exported for use elsewhere in the playbook, then use `certificate_providers`

tests/roles/linux-system-roles.template
```diff
@@ -0,0 +1 @@
+../..
```
richm, line 1: This link should be called `linux-system-roles.certificate` - which is the fully qualified name of your role

tests/tests_default.yml
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Ensure that the role runs with default parameters
+  hosts: all
+
+  roles:
+    - linux-system-roles.template
```
richm, line 7: `linux-system-roles.certificate`

tasks/main.yml
```diff
@@ -0,0 +1,25 @@
...
+    state: present
+  when: "'certmonger' in providers"
+
+- name: Ensure certificate requests
+  certificate_request:
+    name: "{{ item.name | default(omit) }}"
+    dns: "{{ item.dns | default(omit) }}"
+    ca: "{{ item.ca | default(omit) }}"
+    provider: "{{ item.provider | default(omit) }}"
+  loop: "{{ certificate_requests }}"
```
richm, line 25: you should have a `notify: certificate handler to restart services` here

How do you apply the changes?  Do you need to restart the certmonger service to tell it to process the new/changed cert requests?  Is there some other command you use to tell certmonger to process?  Does it matter which provider is used?

That is - you might need a different handler depending on what the provider is - you might not need any handlers at all.

molecule/default/molecule.yml
```diff
@@ -0,0 +1,39 @@
...
+driver:
+  name: docker
+lint:
+  name: yamllint
+  options:
+    config-file: .yamllint.yml
+platforms:
+  - name: centos-6
+    image: docker.io/linuxsystemroles/centos-6
+    privileged: true
```
richm, line 14: is this role supported on centos-6?  If not, you can remove this

we don't yet have a container image for centos-8 support

tasks/main.yml
```diff
@@ -0,0 +1,25 @@
...
+        map(attribute='provider', default=__certificate_provider_default) |
+        unique |
+        list
+      }}
+
+- name: Ensure certmonger is installed
+  package:
+    name: certmonger
+    state: present
+  when: "'certmonger' in providers"
```
richm, line 17: This hasn't been merged yet, but we have best practices for provider-specific tasks and variables: https://github.com/oasis-roles/meta_standards/blob/7fd94fcbb9388b2f28d30b59b1c85d805a9ed062/README.md#supporting-multiple-providers

I suppose you could have two tasks for this - something like this:
```yaml
- name: Load provider specific variables
  # these files are in the role vars/ directory
  include_vars: "{{ __certificate_provider ~ '.yml' }}"
  loop: "{{ __certificate_providers }}"
  loop_control:
    loop_var: __certificate_provider

- name: Perform provider specfic tasks
  # these files are in the role tasks/ directory
  include_tasks: "main-{{ __certificate_provider ~ '.yml' }}"
  loop: "{{ __certificate_providers }}"
  loop_control:
    loop_var: __certificate_provider
```

In `vars/certmonger.yml` you would have
```yaml
__certificate_certmonger_packages:
  - certmonger
__certificate_certmonger_services:
  - certmonger
```

In `tasks/main-certmonger.yml` you would have
```yaml
- name: Ensure certmonger packages are installed
  package:
    name: "{{ __certificate_certmonger_packages }}"
    state: present

- name: Ensure certmonger services are enabled and started
  service:
    name: "{{ item }}"
    state: started
    enabled: yes
  loop: "{{ __certificate_certmonger_services }}"
```

The handler is going to be tricky, since you may need to invoke multiple handlers - something like
```yaml
- name: certificate handler for certmonger
  do something for certmonger
  listen: __certificate_request_handler
  when: 'certmonger' in __certificate_providers

- name: certificate handler for xxx
  do something for xxx provider
  listen: __certificate_request_handler
  when: 'xxx' in __certificate_providers
```
Note the use of `listen: __certificate_request_handler` - this is how you have multiple handlers for a single task - the `certificate_request` task will then use `notify: __certificate_request_handler`

vars/main.yml
```diff
@@ -0,0 +1,5 @@
+# SPDX-License-Identifier: MIT
+---
+# Put the role's internal variables here that are not distribution specific.
+
+__certificate_provider_default: "certmonger"
```
richm, line 5: you don't need double quotes here

vars/default.yml
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: MIT
+---
+# Put internal variables here with values for unrecognized distributions.
+
+# Example:
+__template_packages: []
+__template_services: []
```
richm, line 7: not sure if you will need these - if so, you might need to have different variables for both platform and provider e.g. you might need files like `vars/certmonger_RedHat_7.yml`, `vars/self_signed_Fedora.yml`, etc. - but I wouldn't do that until it becomes necessary - for now, you might only need provider specific files.

richm:
As a general note - since the provider name will be used as an identifier and in filenames, the provider should conform to python identifier naming - only alphanumeric characters and `_` - no `-` - this applies to filenames as well - https://github.com/oasis-roles/meta_standards#naming-things

library/certificate_request.py
```diff
@@ -0,0 +1,119 @@
...
+        """Return a dict with the module arguments."""
+        return dict(
+            name=dict(type="str", required=True),
+            dns=dict(type="list"),
+            ca=dict(type="str"),
+            provider=dict(type="str", default="certmonger"),
+        )
+
+    @property
+    def provider(self):
```
richm, line 98: This isn't really a property, is it?  AFAICT, it will instantiate a new object of class `CertificateRequestXXXProvider` (where XXX is the provider type) every time it is called - that is, it doesn't get/set an object member called `_provider` or some such.

module_utils/certificate/providers/base.py
```diff
@@ -0,0 +1,291 @@
+import os
+
+from abc import ABCMeta, abstractmethod
+
+from cryptography import x509
```
richm, line 5: Is there a dependency on the rpm package that provides this python library?

module_utils/certificate/providers/base.py
```diff
@@ -0,0 +1,291 @@
...
+        """
+        common_name = params.get("common_name")
+        if common_name is None:
+            dns = params.get("dns")
+            if dns:
+                common_name = dns[0]
+
+            ip = params.get("ip")
+            if ip:
+                common_name = ip[0]
```
richm, line 58: which one takes precedence?  that is, if both `dns` and `ip` are defined, which one should `common_name` be assigned to?

module_utils/certificate/providers/base.py
```diff
@@ -0,0 +1,291 @@
...
+
+        return common_name
+
+    @classmethod
+    def load_from_pem(cls, cert_pem=None, csr_pem=None):
+        """Return a CertificateProxy object initialized from PEM data.
+
+        The argument `cert_pem` it's a certificate in PEM format. The argument
+        `csr_pem` it's a CSR in PEM format.
```
richm, line 67: use `is` instead of `it's` here

module_utils/certificate/providers/base.py
```diff
@@ -0,0 +1,291 @@
...
+    def _get_info_from_x509(self, x509_obj):
+        info = {}
+        if not x509_obj:
+            return info
+
+        self._x509_obj = x509_obj
+
+        info["dns"] = self._get_san_values(x509.DNSName)
+        info["ip"] = self._get_san_values(x509.IPAddress)
+        info["common_name"] = self._get_subject_values(NameOID.COMMON_NAME)
```
richm, line 99: Do these `_get_xxx_values` methods return `None` if the field is not present?

module_utils/certificate/providers/base.py
```diff
@@ -0,0 +1,291 @@
...
+        #   Currently each provider is responsible for implementing the CA
+        #   comparison.
+
+        if not isinstance(other, CertificateProxy):
+            raise TypeError(
+                "Cannot compare 'CertificateProxy' with '{}'".format(type(other),)
+            )
+
+        # Remove empty sequencies and strings, false and None values
+        #   before comparison happens
```
richm, line 151: "sequences"

I can see removing empty values like `{}` and `[]` - is it possible that `''` empty string is a significant value?  Is it possible that `false` is a significant value?  What about numeric values? e.g. a field with a value of `0` will not be included.

module_utils/certificate/providers/certmonger.py
```diff
@@ -0,0 +1,142 @@
...
+        If the file is found it's immediately. If not found returns None.
+        """
+        ls_results = os.listdir(self.requests_dir)
+        line_start = "cert_storage_location="
+        for fname in ls_results:
+            file_path = os.path.join(self.requests_dir, fname)
+            fsize = os.stat(file_path).st_size
+            # do not look into file if it's larger than ~50k
+            #   certmonger request files usually avg around 8k
+            if fsize > (50000):
```
richm, line 39: This seems rather arbitrary - also, why `(50000)` instead of `50000`?

module_utils/certificate/providers/certmonger.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # Set Domains
+        for dns in self.csr.dns:
+            command += ["-D", dns]
+
+        if not self.exists_in_certmonger:
+            # Don't attempt to renew when near to expiration
+            command += ["-R"]
+
+        self._run_command(command, check_rc=True)
```
richm, line 141: So this actually requests the cert - I don't think we need a handler, unless we need the handler to check to see if the cert request was successful?  Does `-w` here mean that the cert request task will wait until the certificate and key are issued and placed on disk?  If a cert request requires human interaction e.g. a security officer approving the cert request, might that take hours or so?

module_utils/certificate/providers/certmonger.py
seocam: It is indeed a somehow arbitrary. The point is that if we someone places a big file in this folder the role would open this file looking for information. It could be 10k but in case the user requests a larger certificate the file might be larger then 10k. 50k just gives enough margins  for that.

The parenthesis are there for no reason. I'll remove it.

module_utils/certificate/providers/certmonger.py
seocam: We'll have an option for not waiting. This is just the base structure so I can start work to add the actual supported parameters.

module_utils/certificate/providers/base.py
seocam: That a good question. It's likely that will have cases where None, empty string and False can be meaningful. With the current parameters (just testing parameters) this was not the case.

I was planning to deal with the other cases when I actually find the situation. Is that reasonable?

module_utils/certificate/providers/base.py
seocam: It should be the DNS but the code is actually wrong. I'll fix it.

module_utils/certificate/providers/base.py
richm: > I was planning to deal with the other cases when I actually find the situation. Is that reasonable?

Yes.

module_utils/certificate/providers/base.py
seocam:
```
scampos@cetus:~ $ repoquery --requires --resolve python3-cryptography
Last metadata expiration check: 0:00:05 ago on Fri 15 May 2020 04:05:41 PM -03.
glibc-0:2.31-2.fc32.i686
glibc-0:2.31-2.fc32.x86_64
openssl-libs-1:1.1.1g-1.fc32.i686
openssl-libs-1:1.1.1g-1.fc32.x86_64
python3-0:3.8.2-2.fc32.i686
python3-0:3.8.2-2.fc32.x86_64
python3-cffi-0:1.14.0-1.fc32.x86_64
python3-idna-0:2.8-6.fc32.noarch
python3-six-0:1.14.0-2.fc32.noarch
```

library/certificate_request.py
seocam: That's true. I didn't get very much concerned about that because it should be called just once but that could change in the future. I'll change it.

module_utils/certificate/providers/certmonger.py
```diff
@@ -0,0 +1,142 @@
...
+    def _find_certmonger_request_file(self):
+        """Search for certificate metadata in certmonger data directory.
+
+        Looks for the certificate path, file by file inside certmonger data
+        directory.
+
+        Only open files that are 50k or smaller to avoid getting stuck for
+        a long time.
+
+        If the file is found it's immediately. If not found returns None.
```
richm, line 30: "it's returned immediately"

module_utils/certificate/providers/certmonger.py
richm: Is the problem that `requests_dir` may usually contain lots of different kinds of files, are not discriminated by a file name extension (e.g. .conf), or can have the type determined by looking at the first few 100 or so bytes of the file?  Or is the issue that a malicious user may try a DoS attack by adding a 100GB file to this directory?  I'm just trying to understand the use case here.

module_utils/certificate/providers/certmonger.py
richm: ok - something to consider when developing the handlers

module_utils/certificate/providers/base.py
richm: This code is run on the remote machine - which means it will be using python2 on a rhel7 machine.  I just wanted to make you aware that you may have to do something like - vars/certmonger-RedHat-7.yml
```yaml
__certificate_certmonger_packages: ['certmonger', 'python-cryptography']
```
to ensure the package is available on the remote machine

molecule/default/molecule.yml
seocam: I don't think we'd need centos-6.

tasks/main.yml
seocam: I don't actually think we'll need handlers for now. I'll remove the handlers directory for now.

meta/main.yml
seocam: https://github.com/linux-system-roles/certificate/pull/1/commits/b624c0071a6c1ff291c066cd99df9e4e7a4f1c75

handlers/main.yml
seocam: I've removed handlers https://github.com/linux-system-roles/certificate/pull/1/commits/7012d594bd5ccbbcb7b03ce7bf0922f2ac9425cd

tasks/main.yml
seocam: Handlers removed here: https://github.com/linux-system-roles/certificate/pull/1/commits/7012d594bd5ccbbcb7b03ce7bf0922f2ac9425cd

tasks/main.yml
seocam: 63629e5b05dcf8d1f59a5dc7ca39f9551bc6dfce

tests/roles/linux-system-roles.template
seocam: 4df714a3abd44bbcabce5051b55ccdd76bf1ef5b

tests/tests_default.yml
seocam: 20496808bc647b5c085c961828cc276189a83a02

molecule/default/molecule.yml
seocam: 34ed562fda1139e3b90806c54908fab2d1ac0119

seocam: 34ed562fda1139e3b90806c54908fab2d1ac0119

vars/main.yml
seocam: efa00dad5ca6553db70463757df56dee022a20f6

module_utils/certificate/providers/base.py
seocam: Correct. The python-cryptography will actually be needed by any provider and on centos/rhel 8 the package will be python3-cryptography. I'll fix that with the changes in the `tasks/main.yml`

module_utils/certificate/providers/base.py
seocam: 4e74b07b67c935b8d577375e7779606444905190

module_utils/certificate/providers/base.py
seocam: Typo fixed in 4e74b07b67c935b8d577375e7779606444905190

module_utils/certificate/providers/certmonger.py
seocam: 4e74b07b67c935b8d577375e7779606444905190

module_utils/certificate/providers/certmonger.py
seocam: Originally the directory should only contain text files and the file names do not have extension. The file usually starts with `id=...` but that's not always true. 
I've just checked with @rcritten and he said it should be safe to trust that everything in that directory is a certmonger request file.

library/certificate_request.py
seocam: 99b55cd97725b3448d8cbc9609a900ed94e45486

module_utils/certificate/providers/certmonger.py
seocam: Removed in 0754286301d679dcb2559f2ca6c4f8e9f805776f

module_utils/certificate/providers/base.py
seocam: 0787fbfe8736c210ea1dbd98d75a3d7c4b554933

module_utils/certificate/providers/base.py
```diff
@@ -0,0 +1,291 @@
...
+        self.module.exit_json(
+            changed=self.changed, msg=self.message,
+        )
+
+    def run(self, check_mode=False):
+        """Entry point for the providers called from the actual Ansible module."""
+        if check_mode:
+            self.message += "(Check mode) "
+
+        self.message = "Certificate "
```
rcritten, line 249: This will overwrite the previous (Check mode) if check_mode == True

module_utils/certificate/providers/certmonger.py
```diff
@@ -0,0 +1,136 @@
...
+from .base import CertificateRequestBaseProvider
+
+
+class CertificateRequestCertmongerProvider(CertificateRequestBaseProvider):
+    """Certmonger provider for certificate Linux System Role."""
+
+    def __init__(self, *args, **kwargs):
+        super(CertificateRequestCertmongerProvider, self).__init__(*args, **kwargs)
+
+        self.requests_dir = "/var/lib/certmonger/requests"  # TODO: Allow to override?
```
rcritten, line 17: It's possible it could be different by distribution. I'm not aware that it is different anywhere else though, but the packager may have done it without letting me know (why would they?)

module_utils/certificate/providers/certmonger.py
```diff
@@ -0,0 +1,136 @@
...
+            with open(file_path, "r") as file_:
+                for line in file_:
+                    # find the line with the certificate location
+                    if line.startswith(line_start):
+                        location = line.split("=")[1].strip()
+                        if location == self.certificate_file_path:
+                            return file_path
+                        # not the right file, move to next file
+                        break
+        return None
```
rcritten, line 45: Is this expected to run as root? certmonger is a dbus app so you can query for things over that rather than doing filesystem I/O.

tasks/main.yml
seocam: Added here: 0569e988dba550ec6af56df9c1718422069040ef

vars/default.yml
seocam: Is that really required to split deps in files by provider/distro? 
That's how I've done so far: 0569e988dba550ec6af56df9c1718422069040ef
For global providers I see the point but for task specific providers I can't really see any advantage.

module_utils/certificate/providers/base.py
seocam: f47b48dab848b02f924458f6a59ac75222c7a944c

seocam: 47b48dab848b02f924458f6a59ac75222c7a944c

module_utils/certificate/providers/certmonger.py
seocam: I'll check on Debian-based systems just in case.

module_utils/certificate/providers/certmonger.py
seocam: I decided to go for filesystem because I could easily parse the file using configparser which is a python builtin and to use dbus I'd need to add the bindings as dependency.

By default dbus wouldn't also require root access? Also if I run getcert without root access I would get a msg 'Insufficient access.  Please retry operation as root.'

vars/default.yml
seocam: Basically "__certificate_packages" will vary depending on OS version.

In case of certmonger it's a single service and in case of openssl no service will be required so for now I think we can consider that service doesn't need to be a list.

Shouldn't the defaults.yml and main.yml together contain all variables and the other files (by OS) just overwrite them for doc purpose? I do like when a role is constructed this way because it's really easy to find what was used where.

seocam: New attempt: c85e509de0b98157e5968d72ca56b5b673cad9c8

richm: > 

I think the way it works is that `defaults/main.yml` and `vars/main.yml` are always applied.  Then, you can do something like this https://github.com/linux-system-roles/cockpit/blob/master/tasks/main.yml#L1-L8 to add or replace platform specific variables.

richm: > In case of certmonger it's a single service and in case of openssl no service will be required so for now I think we can consider that service doesn't need to be a list.

Ok, but I think all of the other roles implement support for a list of services, even if there is only 1 service, so this will be inconsistent with other roles.

tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+      {{
+        certificate_requests |
+        map(attribute='provider', default=__certificate_provider_default) |
+        unique |
+        list
+      }}
+
+- name: Ensure provider packages are installed
+  package:
+    name: "{{ __certificate_provider_vars[__certificate_provider].packages }}"
```
richm, line 15: I think you need to add the `__certificate_packages` to this - something like
```
     name: "{{ __certificate_provider_vars[__certificate_provider].packages + __certificate_packages }}"
```

module_utils/certificate/providers/certmonger.py
seocam: Debian-based uses on same location.
TODO removed on 2aeb77af734f4e369b359a41d5d4086f762ee7eb

vars/default.yml
seocam: Alright. I can do same way to keep consistency.

It's just kinda of weird to add code that it's probably not going to be executed anytime soon and will just make the code harder to understand :)

tasks/main.yml
seocam: That's right! :facepalm: 
I will do it in a separe play since this is installing provider specific packages. Otherwise the certificate packages will be installed multiple times (actually the attempt to install will happen multiple times).

vars/default.yml
richm: I think having 1 service is ok - just add a comment to the effect of "yes, I know every other role has a list of services, but this one does not"

seocam: Added. I've done a rebase because there were too many fixups already.

module_utils/certificate/providers/certmonger.py
```diff
@@ -0,0 +1,136 @@
...
+        self._request_file_path = self._find_certmonger_request_file()
+        self._certmonger_metadata = self._read_certmonger_request_file()
+
+    def _find_certmonger_request_file(self):
+        """Search for certificate metadata in certmonger data directory.
+
+        Looks for the certificate path, file by file inside certmonger data
+        directory.
+
+        Only open files that are 50k or smaller to avoid getting stuck for
```
richm, line 32: you can remove this comment (or perhaps you already did in a later commit)

seocam: Nopz! It was lost here. I'll get it removed.

seocam: 36029751c6fcb22a163511a512af21455395304b

richm:
lgtm - would like to get a lgtm from @rcritten and one of the other core system roles team (@pcahyna or @nhosoi)

tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+    loop_var: __certificate_provider
+  when: "__certificate_provider_vars[__certificate_provider].service is defined"
+
+- name: Ensure certificate requests
+  certificate_request:
+    name: "{{ item.name | default(omit) }}"
+    dns: "{{ item.dns | default(omit) }}"
+    ca: "{{ item.ca | default(omit) }}"
+    provider: "{{ item.provider | default(omit) }}"
+  loop: "{{ certificate_requests }}"
```
nhosoi, line 55: Sorry for this novice question. Having default(omit) on all the dict keys, certificate_request accepts, e.g., a request missing name but having the rest?

seocam: Since `certificate_request` is a python module the module implementation is taking care of mandatory/not mandatory and also the defaults. Omitting was just a way to delegate this responsibility to the module.

nhosoi: I see. I.e., I can request a cert with none of the values?

seocam: No. If you don't set any value it will raise an error saying that `name` is required. If you set name if should complain that you didn't set `ca`. If you set `ca` I'm not sure what would happen (I'll test to check).

`ca` and `name` are declared as mandatory in the file `library/certificate_request.py`, more specifically in https://github.com/seocam/certificate-system-role/blob/49aa94bdbd789e0885a2b1620d74b82c95c91b1b/library/certificate_request.py#L93-L100

seocam: @nhosoi it's indeed possible to issue a certificate without setting the dns and it should work (certmonger allows it) but the module code was not working. 

This is the commit with the fix: 875e8c2ef9142105897cba78af13dfd223eff504

nhosoi: Thanks for the clarification, @seocam!

richm:
lgtm

vars/CentOS-7.yml
```diff
@@ -0,0 +1,6 @@
+# SPDX-License-Identifier: MIT
```
pcahyna, line 1: Make this file a symlink to RedHat-7.yml

vars/main.yml
```diff
@@ -0,0 +1,11 @@
+# SPDX-License-Identifier: MIT
+---
+# Put the role's internal variables here that are not distribution specific.
+
+__certificate_provider_default: certmonger
+
+__certificate_provider_vars:
```
pcahyna, line 7: I would suggest to
- not nest variables inside dictionaries - a flatter structure is usually better
- load provider-specific variables from separate files, just as you do for distribution-version.

tasks/main.yml
pcahyna: What do you mean by "separate play"?

richm: > What do you mean by "separate play"?

separate tasks:
```yaml

- name: Ensure certificate role dependencies are installed
  package:
    name: "{{ __certificate_packages }}"
    state: present
...

- name: Ensure provider packages are installed
  package:
    name: "{{ __certificate_provider_vars[__certificate_provider].packages }}"
    state: present
  loop: "{{ __certificate_providers }}"
  loop_control:
    loop_var: __certificate_provider
  when: |
    __certificate_provider_vars[__certificate_provider].packages is defined
```

module_utils/certificate/providers/certmonger.py
pcahyna: Without knowing too much about certmonger, this sounds like a good point - if certmonger has a dbus interface, why doing things behind its back instead of asking it over dbus? The network role also communicates with NetworkManager over dbus.

pcahyna:
Are there unit tests for the Python code in `library` and `module_utils`? I mean not integration tests (Ansible playbooks) but tests written in Python that can run outside Ansible.
Such tests are executed by our template tox/travis setup and have the added benefit of being executed by all Python versions that we need to support (including 2.6, although that may be moot, if you don't intend to support RHEL 6). I believe @richm can advise on them in more detail.

tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+    - "{{ ansible_os_family }}.yml"
+    - "default.yml"
+
+- name: Ensure certificate role dependencies are installed
+  package:
+    name: "{{ __certificate_packages }}"
+    state: present
+
+- name: Find required providers
+  set_fact:
```
pcahyna, line 17: As this variable is used in only a few tasks below, please wrap those tasks in a `block:` and set the variable on a block, instead of making it a fact. (facts are globals and can not be unset, so a local variable is preferred if applicable.)

vars/main.yml
pcahyna: My bad, I did not realize there can be several providers at once - given that, your approach is probably the best one possible.

richm:
No, there are not yet any unit tests.  There aren't any tests at all yet.

seocam:
Correct @richm. @pcahyna there are no tests in this PR yet. The goal here is to have the structure of the role. Next in the PRs I'll start to implement the actual options that it's respectives tests.

We won't need to support RHEL 6.

vars/CentOS-7.yml
seocam: 49792cecbec3979939bc8886b3156f71cf3b006c

tasks/main.yml
seocam: 20269498b23980fda8691698e476bbedac86d65f

module_utils/certificate/providers/certmonger.py
seocam: well... reading a ini-like file seemed a lot more straightforward to me since we are performing only reads in the files. Also by reading the file it prevent a dependency. 

@rcritten what risks do you see in the file implementation? What would be the benefits on using dbus for reading the CSR (pem data) and the ca name from dbus instead of doing from the file?

library/certificate_request.py
```diff
@@ -0,0 +1,127 @@
...
+    dns:
+      - www.example.com
+      - example.com
+    ca: self-sign
+"""
+
+RETURN = ""
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.certificate.providers import certmonger
```
pcahyna, line 66: It looks that most of the actual logic is under `module_utils.certificate.providers`. Would it make sense to have one module per provider, instead of having a single module that dispatches to the appropriate provider? That way you would not need to hardcode specific providers into common code.

seocam: Most of the logic is in `providers.base`. Even if we have a module per provider I think that would basically remains the same.

I get your point about how the coupling was designed... 
We could use a constant in the providers.__init__ and just import this constant here. What do you think?

seocam: Do you think this way would be better? fa8a7981490bf0ceb24f02961238f861705940c5

module_utils/certificate/providers/certmonger.py
rcritten: There could be a race condition between the state inside certmonger and the state on disk. This isn't theoretical, it was seen in IPA and at first mitigated (retries) and then switched to dbus.

seocam: That's a very strong point. In this case using dbus for quering the information and the cli for requesting the certs would mitigate that?

library/certificate_request.py
```diff
@@ -38,6 +38,12 @@
         certificate.
     required: false
     default: certmonger
+  directory:
+    description:
+      - Directory where certificate and key will be stored. Only used
+        if I(name) is not a fully qualified path.
```
pcahyna, line 44: I would say "absolute path" instead of "fully qualified path"

seocam: Amended.

pcahyna: By the way, what will happen if `name` does not start with a slash, but it contains slash(es)? Will it put the certificate file under a subdirectory of `directory`?

seocam: Correct. The paths are joined using `os.path.join`.

pcahyna: if `name is `foo/bar` and there is no `/etc/pki/tls/certs/foo` directory, will it be created? I don't see it in the source, but maybe that would be certmonger's job.
By the way, in case of subdirectories, should the prefix be `/etc/pki/tls/certs/foo` or `/etc/pki/tls/foo/certs`?
The `certs` and `private` subdirectories are not distribution-specific?

seocam: @pcahyna good points!

Although `certs` and `private` directories could probably be considered distro specific they are used inside the default path of pretty much everything I found: OpenBSD, Slack, Alpine, Gentoo and deb-based systems.

In case of sub-directories /etc/pki/ should be the correct base directory for rpm-based systems. For all the other cases /etc/ssl should not contain sub-directories.
Because of that I'm thinking that name should be either an absolute path or a simple string and the module should raise an error if a non-abspath is given.

Does that make sense?

pcahyna: Is it usual to have user- or package- defined subdirectories under `/etc/pki/`? If so, are they directly under `/etc/pki` or under `/etc/pki/tls`? If they are not expected, it makes indeed the best sense to forbid them.

pcahyna: In the case of absolute path, no `certs`/`private` path components are inserted. Is it intended and expected?

seocam: On RPM-based there are package defined sub-directories under `/etc/pki` (without `tls`).

In that case I'm going to perform the changes to only allow simple string or abspaths.

pcahyna: So if anyone needs subdirectories, they should be implemented as a separate option instead.

vars/Fedora.yml
```diff
@@ -0,0 +1,3 @@
+# SPDX-License-Identifier: MIT
+---
+# Put internal variables here with Fedora specific values.
```
pcahyna, line 3: `__certificate_packages` is missing here. Also probably `__certificate_default_directory: /etc/pki/tls`

vars/RedHat-7.yml
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: MIT
+---
+# Put internal variables here with Red Hat Enterprise Linux 7 specific values.
+
+__certificate_packages:
+  - python-cryptography
+  - python-dbus
```
pcahyna, line 7: `__certificate_default_directory: /etc/pki/tls` is missing here, I think.

vars/Fedora.yml
seocam: I think we can remove this file since the vars will be loaded from defaults.

seocam: Amended.

vars/RedHat-7.yml
seocam: Amended.

library/certificate_request.py
seocam: Amended.

rcritten:
The certmonger dbus code looks good. Did some unit testing on it and it works fine.

library/certificate_request.py
```diff
@@ -38,6 +39,12 @@
         certificate.
     required: false
     default: certmonger
+  directory:
+    description:
+      - Directory where certificate and key will be stored. Only used
+        if I(name) is not a absolute path.
```
pcahyna, line 45: "an absolute path"

seocam: Amended.

vars/Debian-10.yml
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: MIT
```
pcahyna, line 1: would it make sense to name the file just "Debian", or are there relevant changes between Debian versions, or expected in the future?

.github/CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+# https://help.github.com/en/articles/about-code-owners
+# Default reviewers for everything
+* @i386x @pcahyna @richm
```
pcahyna, line 3: @seocam you should add yourself here and remove i386x

seocam: ok!

vars/Debian-10.yml
seocam: There are differences in python version in package name "python3" or just "python" (similar to Centos versions).

.github/CODEOWNERS
seocam: https://github.com/linux-system-roles/certificate/pull/1/commits/7369b739334fbdc60be814795b2dbbffae51533b

richm:
other than one minor nit, lgtm
not sure what's up the the travis CI status

vars/Debian-10.yml
pcahyna: Wait, if there is no Debian.yml, where is `__certificate_default_directory: /etc/ssl` going to be defined for other Debians?

seocam:
> Can you please address the question about Debian?

Right. This way it's only supporting Debian 10 (which is actually the only one I've tested with). I thought that if people is interested in running on other distros they could do the PR for it.

I just added Debian 10 to have at least one example of non-rpm-based distro.

vars/Debian-10.yml
pcahyna: > There are differences in python version in package name "python3" or just "python" (similar to Centos versions).

Which version of Debian has `python3` in the names of the required packages? Debian 10 is the latest as of now. It it Debian testing that transitions to `python3` naming? Or were those the older versions that used `python3` and the newer ones transitioned to just `python`?

seocam: On debian buster (10) both packages are available: 
https://packages.debian.org/buster/python-cryptography
https://packages.debian.org/buster/python3-cryptography

Running ansible on a cloud image on Debian 10, by default is using python2 and if I install python3-cryptography the execution just fails.

I haven't tested in any other Debian versions because Debian is actually out-of-scope for what we need to finish by June 15th. Do you think it's reasonable that we open an issue to add support to other Debian/Ubuntu versions and take this discussion there?

pcahyna: Sure, you don't need to block it on this issue. i am just pointing out that if there are no known differences of packages required between Debian versions, then your comment
> There are differences in python version in package name "python3" or just "python" (similar to Centos versions).

is not accurate.

seocam: I see. By having `Debian.yml` gives a chance that it might work in other Debian versions (even if not tested) while having `Debian-10.yml` enforces that it will work only on version 10.

I've renamed the file and will open an issue for testing that.

pcahyna: It does not even enforce it, because on non-10 Debian it will pick `default.yml` and do the wrong thing without warning.

pcahyna: (you probably meant a different thing by "enforces".)

PR #3, opened 2020-05-26
Initial integration tests

seocam:
Adds initial integration tests.

### DIFF
--- a/tests/tasks/assert_certificate_parameters.yml
+++ b/tests/tasks/assert_certificate_parameters.yml
@@ -1,0 +1,53 @@
+---
+- name: Retrieve certificate file stats
+  stat:
+    path: "{{ cert['path'] }}"
+  register: result
+
+- name: Verify if certificate file exists
+  assert:
+    that:
+      - result.stat.exists
+    fail_msg: "Certificate file '{{ cert['path'] }}' does not exist."
+
+- name: Retrieve key file stats
+  stat:
+    path: "{{ cert['key_path'] }}"
+  register: result
+
+- name: Verify if key file exists
+  assert:
+    that:
+      - result.stat.exists
+    fail_msg: "Key file '{{ cert['key_path'] }}' does not exist."
+
+- name: Retrieve certificate subject
+  shell: >-
+    openssl x509 -in {{ cert['path'] }} -noout -subject |
+    sed 's/ = /=/g' |
+    sed 's/^subject=\(.*\)$/\1/'
+  register: result
+  changed_when: false
+
+- name: Verify certificate subject
+  assert:
+    that:
+      - cert['subject'] == result.stdout
+    fail_msg: "{{ cert['subject'] }} != {{ result.stdout }}"
+
+- name: Retrieve certificate Subject Alternative Name (SAN)
+  shell: >-
+    openssl x509 -in {{ cert['path'] }} -noout -text |
+    grep 'Subject Alternative Name' -A1 |
+    tail -1 |
+    tr , '\n' |
+    sed 's/^\s\+//g'
+  register: result
+  changed_when: false
+
+- name: Verify certificate SAN
+  assert:
+    that:
+      - (cert['subject_alt_name'] | sort) == (result.stdout_lines | sort)
+    fail_msg: >-
+      {{ cert['subject_alt_name'] | sort }} != {{ result.stdout_lines | sort }}
--- a/tests/tests_wrong_provider.yml
+++ b/tests/tests_wrong_provider.yml
@@ -1,0 +1,26 @@
+---
+- name: Test issuing certificate with nonexistent provider
+  hosts: all
+  become: true
+  vars:
+    certificate_requests:
+      - name: mycert
+        dns: www.example.com
+        ca: self-sign
+        provider: fake-provider
+
+  tasks:
+    - vars:
+        expected_error_msg: >-
+          Ensure certificate requests Chosen provider
+           'fake-provider' is not available.
+      block:
+        - import_role:
+            name: linux-system-roles.certificate
+        - fail:
+            msg: "Certificate issued with nonexistent provider 'fake-provider'."
+      rescue:
+        - name:
+          assert:
+            that:
+              - ansible_failed_task.name == "Ensure certificate requests"
--- a/examples/basic_self_signed.yml
+++ b/examples/basic_self_signed.yml
@@ -9,4 +9,4 @@
         ca: self-sign
 
   roles:
-    - certificate
+    - linux-system-roles.certificate
--- a/examples/many_self_signed.yml
+++ b/examples/many_self_signed.yml
@@ -8,11 +8,11 @@
         dns: www.example.com
         ca: self-sign
       - name: /tmp/other-cert
-        dns: www.example.com
+        dns: www.example.org
         ca: self-sign
       - name: /tmp/another-cert
-        dns: www.example.com
+        dns: www.example.net
         ca: self-sign
 
   roles:
-    - certificate
+    - linux-system-roles.certificate
--- a/examples/wrong_provider.yml
+++ b/examples/wrong_provider.yml
@@ -10,4 +10,4 @@
         provider: fake-provider
 
   roles:
-    - certificate
+    - linux-system-roles.certificate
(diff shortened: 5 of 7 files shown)


seocam:
Needs to be rebased after #1 is merged.

seocam:
@amore17 if you look at the last commit you will see a few examples of tests.

richm:
initial review looks good - will review again once #1 is merged and this pr is rebased

seocam:
CentOS/RHEL 7 is failing because of: https://github.com/linux-system-roles/certificate/issues/5

Since the error was not introduced by this PR a propose that the we work in the fix in a different PR.

tests/tasks/assert_certificate_parameters.yml
```diff
@@ -0,0 +1,35 @@
+---
+- name: Verify certificate exists in the correct location
+  file:
+    path: "{{ cert['path'] }}"
+    state: file
+  register: result
```
richm, line 6: do you need the register here?

tests/tasks/assert_certificate_parameters.yml
```diff
@@ -0,0 +1,35 @@
...
+
+- name: Verify key exists in the correct location
+  file:
+    path: "{{ cert['key_path'] }}"
+    state: file
+  register: result
```
richm, line 12: do you need the register here?

tests/tasks/assert_certificate_parameters.yml
richm: did you mean to use `stat` instead of `file`?  Something that would throw an error if the file does not exist?

tests/tasks/assert_certificate_parameters.yml
```diff
@@ -0,0 +1,35 @@
...
+
+- name: Verify certificate subject
+  shell: >-
+    openssl x509 -in {{ cert['path'] }} -noout -subject |
```
richm, line 16: Maybe this test could check for the existence of `cert['path']` as well as testing the contents of the cert.  That is, I think `openssl x509 -in file` will fail if `file` does not exist.  Unless you need to distinguish between "file does not exist" and "file contents are not correct"

tests/tasks/assert_certificate_parameters.yml
seocam: Actually I wasn't sure at all about what to use here. In other PR I've used file and registered the result to verify with assert. It seems like only `stat` would be able to do the job by itself, right?
I'll updated it.

tests/tasks/assert_certificate_parameters.yml
seocam: Same then https://github.com/linux-system-roles/certificate/pull/3#discussion_r433869562

tests/tasks/assert_certificate_parameters.yml
seocam: The idea is that if the file does not exist the error would happen in some assertion before we get here.

tests/tasks/assert_certificate_parameters.yml
seocam: `stat` only added the "assertion" capabilities starting from ansible 2.9.

seocam: Using `fail` module is preferable then using `assert` module?
`is exists` is preferable over `stat.exists`. Note that we still need stat to check owner/group in next PRs.

richm: > Using `fail` module is preferable then using `assert` module?

No.  Either is fine.

> `is exists` is preferable over `stat.exists`. Note that we still need stat to check owner/group in next PRs.

Ah, if you need the result of `stat` then keep it.

nhosoi: I also have register in the stat task in the logging role CI-tests. Should I also consider getting rid of the register?
Regarding the ansible version, I see 2.9 is available on RHEL-7 and 8 as well as Fedora's. Do we still have to worry about older versions of ansible?

tests/tasks/assert_certificate_parameters.yml
```diff
@@ -1,35 +1,53 @@
...
     path: "{{ cert['path'] }}"
-    state: file
   register: result
 
-- name: Verify key exists in the correct location
-  file:
+- name: Verify if certificate file exists
+  assert:
+    that:
+      - result.stat.exists == true
```
nhosoi, line 10: In the similar code, I was told `== true` is redundant. ;)

richm: right - and to avoid the "bare variable" problem, use this: `result.stat.exists | bool`

seocam: That's `true` :sweat_smile:
I'll get it fixed.

tests/tasks/assert_certificate_parameters.yml
richm: @seocam do you have a link that says "assert" is only supported in `stat` in ansible 2.9 and later?

seocam: As a general rule in Red Hat we currently should support 2.7+ (in a few weeks only 2.8+)[1]. I'm not sure if that applies to linux-system-roles as well.

[1] https://access.redhat.com/support/policy/updates/ansible-engine

tests/tasks/assert_certificate_parameters.yml
seocam: Fixed in: f1bbf8139c67d180856bceb508c72ea86293e14a

tests/tasks/assert_certificate_parameters.yml
richm: > As a general rule in Red Hat we currently should support 2.7+ (in a few weeks only 2.8+)[1]. I'm not sure if that applies to linux-system-roles as well.

It applies - we support ansible 2.7, 2.8 and 2.9
.
> 
> [1] https://access.redhat.com/support/policy/updates/ansible-engine

seocam:
If you are ok with the changes until here I'll squash them.

tests/tests_wrong_provider.yml
```diff
@@ -0,0 +1,26 @@
...
+      block:
+        - import_role:
+            name: linux-system-roles.certificate
+        - fail:
+            msg: "Certificate issued with nonexistent provider 'fake-provider'."
+      rescue:
+        - name:
+          assert:
+            that:
+              - ansible_failed_task.name == "Ensure certificate requests"
```
pcahyna, line 26: It is better to avoid `ansible_failed_task`, because it can lead to unexpected failures. See https://github.com/linux-system-roles/storage/pull/88

tests/tasks/assert_certificate_parameters.yml
pcahyna: @richm I think that `result.stat.exists` is not a bare variable (that would be just `result`).

tests/tasks/assert_certificate_parameters.yml
pcahyna: @seocam what are those "assertion" capabilities starting from ansible 2.9? I do not see any assertion mentioned in https://docs.ansible.com/ansible/latest/modules/stat_module.html

pcahyna: > in a few weeks only 2.8+

That's weird, I have supposed that 2.7 will be supported until the release of 2.10.

pcahyna: > `is exists` is preferable over `stat.exists`

no, `is exists` checks for paths on the controller and is totally unrelated to the return value of `stat`. https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#testing-paths

tests/tests_wrong_provider.yml
seocam: I've addressed that in the "next" PR https://github.com/linux-system-roles/certificate/pull/6

tests/tasks/assert_certificate_parameters.yml
pcahyna: > > in a few weeks only 2.8+
> 
> That's weird, I have supposed that 2.7 will be supported until the release of 2.10.

I asked @maxamillion for clarification.

PR #6, opened 2020-06-01
Support to DNS, IP and Email SAN fields

seocam:


seocam:
This PR depends on #3 but the commit `Support to DNS, IP and Email SAN fields` can be reviewed.

examples/basic_self_signed.yml
```diff
@@ -9,4 +9,4 @@
         ca: self-sign
 
   roles:
-    - certificate
+    - linux-system-roles.certificate
```
richm, line 12: This change looks unrelated to the subject of the PR, unless it is necessary to run the test

module_utils/certificate/providers/base.py
```diff
@@ -36,10 +37,14 @@ def load_from_params(cls, module, params):
         # pylint: disable=protected-access
         cert_like = cls(module)
 
-        map_attrs = ["dns", "ip"]
+        map_attrs = ["dns", "ip", "email"]
         info = {k: v for k, v in params.items() if k in map_attrs}
 
         info["common_name"] = cert_like._get_common_name_from_params(params)
+        if info.get("ip"):
```
richm, line 44: this works, but the typical python idiom is to use `if "ip" in info` since you are not otherwise using the result of `get()`

module_utils/certificate/providers/certmonger.py
```diff
@@ -136,6 +136,18 @@ def request_certificate(self):
         # Set Domains
         for dns in self.csr.dns:
             command += ["-D", dns]
+        if not self.csr.dns:
+            command += ["-D", ""]
```
richm, line 140: this seems odd - why wouldn't you just omit the `-D` argument?  Or does that have a different semantic meaning than `-D ""`?

tests/tests_wrong_provider.yml
```diff
@@ -0,0 +1,26 @@
...
+      block:
+        - import_role:
+            name: linux-system-roles.certificate
+        - fail:
+            msg: "Certificate issued with nonexistent provider 'fake-provider'."
+      rescue:
+        - name:
+          assert:
+            that:
+              - ansible_failed_task.name == "Ensure certificate requests"
```
richm, line 26: `ansible_failed_task` is currently broken - https://github.com/ansible/ansible/issues/57399

See https://github.com/linux-system-roles/logging/blob/master/tests/tests_files_elasticsearch_certs_incomplete.yml#L24 for the current best practices idiom of testing for a failure with `block/rescue` by using a dummy `name: unreachable task` and `ansible_failed_result`

examples/basic_self_signed.yml
seocam: This is actually part of PR #3 but it's also not required by that PR itself. It's there because running the tests I'm running the examples manually to ensure they work.

module_utils/certificate/providers/certmonger.py
seocam: It's indeed weird but it's required. 

For example, if we have 2 runs with the following settings:
First:
```
    certificate_requests:                                                                                 
      - name: mycert                                                                                      
        dns:                                                                                              
          - www.example.com                                                                                                                                    
        ca: self-sign    
```
Second:
```                                                                                
    certificate_requests:                                                                                 
      - name: mycert                                                                                      
        ip:                                                                                               
          - 192.0.2.12                                                                                         
        ca: self-sign    
```

If I don't set `-D ""` certmonger will leave `www.example.com` in the certificate which I don't think it's the behavior we want the role to have.

module_utils/certificate/providers/base.py
seocam: I actually want to get the value and verify that's not `None` neither `[]`.

richm: ah, ok - then what you have is shorter and more readable then `if "ip" in info and info["ip"]`

seocam: @richm alright. Although actually a little longer the intention is obviously a lot clearer in that form. I'll change.

richm: I think you should leave it as `if info.get("ip"):` - I think that is shorter and more readable than the alternative (imo)

seocam: I think that if I had `if "ip" in info and info["ip"]` you probably wouldn't have the doubt you had. That's why I assumed the intention would be clearer in the longer form.
But this is so small that I'm ok with either way.

tests/tests_wrong_provider.yml
seocam: Updated in b8a4c36c7b30d2da633b06937ce862c674cc0c91

module_utils/certificate/providers/certmonger.py
```diff
@@ -110,6 +110,7 @@ def exists_in_certmonger(self):
 
     def request_certificate(self):
         """Issue or update a certificate using certmonger."""
+        # pylint: disable=useless-else-on-loop
```
richm, line 113: Why is the `else` useless?  Does the linter think that e.g. `self.csr.dns` will always have at least 1 element?

seocam: `pylint` complains that there is an `else` in a `for` loop when there is no `break` inside the loop. The `for-else` was designed for fallback in cases where the programmer is looking for something in the loop and can't find it.
Since we are using it in a different way we need to disable this warning.

Initially I've used `if not` to avoid the lint complains without having to explicitly silent it.

richm: Ok - I see - `for/else` is supposed to be used with a `break` - I have typically used it without a break, but not in linted code . . . then I'll leave it up to you if you want to revert back to the `if not xxx` version, or keep the current code with the pylint suppression.

seocam: Ok. I fine with the current way.

richm:
lgtm

pcahyna:
I believe you should omit all those `become: yes` in tests, I don't think we use them elsewhere. In general, our roles do not make much sense to execute as other user than root or without become.

amore17:
> I believe you should omit all those `become: yes` in tests, I don't think we use them elsewhere. In general, our roles do not make much sense to execute as other user than root or without become.
IMO we can keep `become: yes` as it is as these tests will be running in downstream IDM-CI also.

PR #7, opened 2020-06-01
Support to set owner/group for cert and key

seocam:


seocam:
This PR depends on #6 but the commit `Support to set owner/group for cert and key` can be reviewed.

library/certificate_request.py
```diff
@@ -26,7 +26,31 @@
...
+  email:
+    description:
+      - Email (or list of emails) to be included in the
+        certificate. Also can provide the default value for
+        I(common_name).
+    required: false
+  owner:
+    description:
+      - User name (or user id) for the certificate and key files.
+    required: false
```
richm, line 46: Is there a default value for `owner` and `group` e.g. `UMASK` or such ?

module_utils/certificate/providers/base.py
```diff
@@ -267,26 +283,53 @@ def _exit_success(self):
...
-                # If the cert exists and does not require updates we are done
-                self._exit_success()
 
-        if check_mode:
-            self._exit_success()
+        if issue_or_update_cert and not check_mode:
+            self.request_certificate()
 
-        self.request_certificate()
+        updated_fs_attrs = self._set_user_and_group_if_different()
```
richm, line 330: What happens if `check_mode` is true and `self._set_user_and_group_if_different()` is true?  Does that mean the system state changed when using `check_mode`?

seocam: This method calls the method from the basic module implementation (https://github.com/ansible/ansible/blob/v2.8.0/lib/ansible/module_utils/basic.py#L1327) which already takes care of `check_mode`.

library/certificate_request.py
seocam: Not really. I was leaving that to Certmonger default.

module_utils/certificate/providers/certmonger.py
```diff
@@ -136,10 +136,23 @@ def request_certificate(self):
         # Set Domains
         for dns in self.csr.dns:
             command += ["-D", dns]
+        if not self.csr.dns:
```
richm, line 139: use `else` instead of a separate `if not` clause.  also below

tests/tasks/assert_certificate_parameters.yml
```diff
@@ -10,6 +10,16 @@
       - result.stat.exists
     fail_msg: "Certificate file '{{ cert['path'] }}' does not exist."
 
+- name: Verify certificate file owner and group
+  assert:
+    that:
+      - result.stat.pw_name == (cert['owner'] | default('root'))
```
richm, line 16: `User name (or user id) for the certificate and key files` - this test will fail if `owner` was given as a user id.  You should add a comment in the test to warn future test writers.

PR #8, opened 2020-06-01
Allow to set key size

seocam:


seocam:
Depends on #3

module_utils/certificate/providers/certmonger.py
```diff
@@ -141,5 +141,8 @@ def request_certificate(self):
             # Don't attempt to renew when near to expiration
             command += ["-R"]
 
+        # Set certificate key size
+        command += ["-g", str(self.module.params.get("key_size"))]
```
richm, line 145: Is `-g` a required argument?

library/certificate_request.py
```diff
@@ -28,6 +28,11 @@
         certificate. Also can provide the default value for
         common_name.
     required: false
+  key_size:
```
richm, line 31: This should also be documented in the `README.md` for the role parameters - this is just for the module parameters.

module_utils/certificate/providers/certmonger.py
seocam: It's not required but I'm setting it always to ensure `key_size` will have the default value we chose (currently 2048).

library/certificate_request.py
seocam: I was planning to add the README from our specs as the start point. I just didn't add yet because we are actually missing most of what's there.

tests/tasks/assert_certificate_parameters.yml
```diff
@@ -51,3 +51,20 @@
...
+  changed_when: false
+  when: cert.get("key_size")
+
+- name: Verify key file attributes
+  assert:
+    that:
+      - (result.stdout | int) == (cert["key_size"] | int)
+    fail_msg: |
+      key_size {{ result.stdout }} != {{ cert['key_size'] | default(2048) }}
+  when: cert.get("key_size")
```
richm, line 70: Is it necessary to check that the key_size is 2048, even if `cert.key_size` was not specified?

richm: so you don't need the `when: cert.get("key_size")` here? or is `cert.key_size` always set?

seocam: The `when` is there by mistake.

pcahyna:
Shouldn't it respect crypto policy (i.e., don't create 2048 bit RSA certs if the crypto policy disallows them)?

PR #9, opened 2020-06-01
Support to Kerberos principal

seocam:


library/certificate_request.py
```diff
@@ -100,6 +112,7 @@ def _get_argument_spec():
             ca=dict(type="str", required=True),
             directory=dict(type="str", default="/etc/pki/tls"),
             provider=dict(type="str", default="certmonger"),
+            principal=dict(type="list"),
```
richm, line 115: I probably should have asked this earlier, but when you declare a module parameter as type `list`, then pass a string to ti as `principal: HTTP/www.example.com@EXAMPLE.com` - what happens?  Is it used in the module as a list with 1 string element?

seocam: Exactly. The module takes care of changing it to a list with a single string on it.

module_utils/certificate/providers/base.py
```diff
@@ -8,12 +8,78 @@
...
+            six.ensure_text(str(n))
+            .replace("\\", "\\\\")
+            .replace("/", "\\/")
+            .replace("@", "\\@")
+            for n in name
+        )
+        name = u"%s@%s" % (name, realm)
+        return name
+
+
```
richm, line 82: @rcritten - does this look ok to you?

seocam:
This PR doesn't have tests yet because OpenSSL does not support displaying Kerberos principals. I'll have to find a another way to print the principal in order to test.

Even in Python is not quite straightforward to see it (as you can see in the PR).

module_utils/certificate/providers/base.py
seocam: It might be good if @tiran can also take a look.

module_utils/certificate/providers/certmonger.py
```diff
@@ -141,5 +141,11 @@ def request_certificate(self):
             # Don't attempt to renew when near to expiration
             command += ["-R"]
 
+        # Set Kerberos principal
+        for principal in self.csr.principal:
+            command += ["-K", principal]
```
richm, line 146: do you need to `shlex.quote` this string?

seocam: That's a good question. It might be safer adding it not only here but in other places as well (line `common_name`.

module_utils/certificate/providers/base.py
rcritten: IPA has a class for handling principals. I'm not sure if this code is based on that or home grown. It would be nice if the IPA code could be used.

seocam: @rcritten the implementation it's heavily based on IPA x509 code.

seocam: We could just install and use ipalib.x509 instead of implementing this here but that will add 126 package dependencies to linux-system-roles [1]. Since the code is not that large I decided to go for the local implementation. @richm @pcahyna what do you think?

[1] https://gist.github.com/seocam/a1bb35dbc68feff90ea08a565d35f9c1

richm: If using the `certmonger` provider, will the system already have the `ipalib.x509` package installed?  If not, will installing it pull in a lot of unnecessary stuff?

seocam: No, `certmonger` does not depend on `ipalib`.

richm: I'm ok with this copy/paste implementation, as it is a standard and unlikely to change, and doesn't seem to be particularly security sensitive.  If someone has a strong objection, please comment.

seocam: 19 extra packages on CentOS 8.

richm: > 19 extra packages on CentOS 8.

Doesn't seem worth it

rcritten: +1, it's ok with me.

module_utils/certificate/providers/certmonger.py
seocam: Using `shlex.quote` is actually adding undesired quotes to the result. For example:
```
"CN=My Certificate with SAN != CN='My Certificate with SAN'"
```
I'm calling `run_command` which will call `Popen` using the list as the command. It seems quoting is not required when using `Popen` this way. It would be required in case I was using a `str` command and `shell=True`.

richm: ok - as long as it is shell safe

seocam:
Pylint failed because of a false-positive: https://travis-ci.com/github/linux-system-roles/certificate/jobs/343282623
The false positive bug still happens but mostly on Python2: https://github.com/PyCQA/pylint/issues/1498
I'll silence this error in that method.

richm:
lgtm

PR #10, opened 2020-06-02
Support to key usage and extended key usage

seocam:


library/certificate_request.py
```diff
@@ -61,6 +61,31 @@
...
+      - decipherOnly
+    required: false
+    default:
+      - digitalSignature
+      - keyEncipherment
+  extended_key_usage:
+    description:
+      - Extended Key Usage attributes to be present in the
+        certificate request.
+    required: false
```
richm, line 85: did you want to list the `choices` here too?  Or can this field take arbitrary values other than the ones listed in `EXTENDED_KEY_USAGE_CHOICES`?

seocam: Initially I've added `EXTENDED_KEY_USAGE_CHOICES` but then I remembered that during the API design @rcritten explained me that EKU can be any OID so we shouldn't limit it to a small list of known aliases.

I removed the constant.

library/certificate_request.py
```diff
@@ -184,6 +246,10 @@ def _get_argument_spec():
             owner=dict(type="str"),
             group=dict(type="str"),
             principal=dict(type="list"),
+            key_usage=dict(
```
richm, line 249: If I wanted to create a cert with no key_usage and no extended_key_usage, would I do this?
```yaml
key_usage: []
extended_key_usage: []
```
?

seocam: You can but with some limitations. Certmonger can issue certificates without `keu_usage` and `extended_key_usage` but it cannot remove all `key_usage` from a certificate.

While this is a limitation in Certmonger it's not really a big problem because there is not much use for a certificate with no `key_usage`. 

@rcritten does that sound right?

rcritten: Agreed. I don't know if it is even valid per the RFC(s) to have no key_usage but it would be a very silly thing to do. I don't know what the use-case would be.

PR #12, opened 2020-06-03
Allow not waiting for certificate to be available

seocam:


pcahyna:
Can it be somehow integrated with asynchronous actions in Ansible? https://docs.ansible.com/ansible/latest/user_guide/playbooks_async.html I never used that feature myself.

seocam:
> Can it be somehow integrated with asynchronous actions in Ansible? https://docs.ansible.com/ansible/latest/user_guide/playbooks_async.html I never used that feature myself.

I haven't heard about this before. I'll take a look.

seocam:
I'm looking into the failing test.

seocam:
> I'm looking into the failing test.

We cannot set owner/group synchronously when asking to not wait for certificate because the file won't exist yet. As we discussed on IRC we'll have to create a script an ask for certmonger to run this just after issuing the certificate. 

As pre/post scripts will be implemented in one of the future PRs I've added a fail message for when users try to combine `wait=False` with `owner` or `group` options. 

https://github.com/linux-system-roles/certificate/blob/aa292cef6b96248dfd7d585c5b30607672e9fa48/module_utils/certificate/providers/certmonger.py#L110-L123

This will be removed as soon as we have the `post run` capability.

richm:
lgtm - but the failures on el7 don't look good

PR #20, opened 2020-06-05
Added support to IPA as CA for Certmonger

seocam:


seocam:
We need `ipaserver` and `ipaclient` roles from `ansible-freeipa` collection in order to run the tests.

richm:
> We need `ipaserver` and `ipaclient` roles from `ansible-freeipa` collection in order to run the tests.

I think that's fine - @sergio-correia did something similar for the nbde_client test, by using the nbde_server role to install a test server to use for testing.  You can see how he did it at https://github.com/linux-system-roles/nbde_client/pull/3/files#diff-4acbded5dcbfc849fef31126ed04b890R15

seocam:
> > We need `ipaserver` and `ipaclient` roles from `ansible-freeipa` collection in order to run the tests.
> 
> I think that's fine - @sergio-correia did something similar for the nbde_client test, by using the nbde_server role to install a test server to use for testing. You can see how he did it at https://github.com/linux-system-roles/nbde_client/pull/3/files#diff-4acbded5dcbfc849fef31126ed04b890R15

But the role has to be installed in the controller, right? I'd have to perform a run local in this case?

I was thinking about adding a .sh script and ask for the executing that in the CI before running the tests (maybe there is already something like that available).

sergio-correia:
> > We need `ipaserver` and `ipaclient` roles from `ansible-freeipa` collection in order to run the tests.
> 
> I think that's fine - @sergio-correia did something similar for the nbde_client test, by using the nbde_server role to install a test server to use for testing. You can see how he did it at https://github.com/linux-system-roles/nbde_client/pull/3/files#diff-4acbded5dcbfc849fef31126ed04b890R15

Actually I wasn't using the nbde_server role yet, but instead had created a couple of tasks to install the package and start the service. I am now (not in the PR yet) using the nbde_server as a git submodule in tests/roles, so that I don't reinvent the wheel.

richm:
For upstream testing, I think it would be ok to have the test playbook use `command` or `shell` to run a script to download and run the roles from galaxy or github.  But we can't do that for downstream testing.

seocam:
> For upstream testing, I think it would be ok to have the test playbook use `command` or `shell` to run a script to download and run the roles from galaxy or github. But we can't do that for downstream testing.

But we'd need to run this is a `local_action`, correct?

richm:
> > For upstream testing, I think it would be ok to have the test playbook use `command` or `shell` to run a script to download and run the roles from galaxy or github. But we can't do that for downstream testing.
> 
> But we'd need to run this is a `local_action`, correct?

Yes e.g. delegate to localhost or something like that.

module_utils/certificate/providers/certmonger.py
```diff
@@ -53,6 +53,35 @@ def __init__(self, *args, **kwargs):
...
 
+    def module_input_validation(self):
+        """Validate module input."""
+        principal = self.module.params.get("principal")
+        ca = self.module.params.get("ca")
+
+        # 'IPA' CA requires principal to be set
+        if ca.lower() == "ipa" and not principal:
+            self.module.fail_json(
+                msg=("Principal parameters is mandatory for 'ipa' CA.")
```
richm, line 64: parameter

richm:
Yes, I think this is the right direction

seocam:
For some reason IPA installation is not completing when running on CI so I've spitted this PR in 2: 
* Feature PR (this one)
* Test PR (#23)

PR #22, opened 2020-06-08
Integration Tests

amore17:
1: Added test with UID and GID for owner and group.
2: Added test to check default auto-renew is set or not.

### DIFF
--- a/tests/tests_fs_attrs.yml
+++ b/tests/tests_fs_attrs.yml
@@ -1,4 +1,16 @@
 ---
+- name: Ensure UID and GID exists
+  hosts: all
+  become: true
+  tasks:
+    - name: Ensure user exists
+      user:
+        name: user1
+        uid: 1040
+    - name: Ensure group "somegroup" exists
+      group:
+        name: somegroup
+        gid: 1041
 - name: Issue certificate setting user/group
   hosts: all
   become: true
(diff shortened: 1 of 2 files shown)


pcahyna:
[citest commit:6cfcd003514ad65b2b7e762f4dc7903f891d2521]

pcahyna:
@amore17 @seocam are those test failures expected?

tests/tasks/presetup.yml
```diff
@@ -0,0 +1,9 @@
+---
```
seocam, line 1: Is this something we are going to reuse in multiple playbooks?
If we are only going to use on `tests/tests_fs_attrs.yml` I would just define it there.

tests/tests_fs_attrs.yml
```diff
@@ -9,6 +16,12 @@
         owner: ftp
         group: ftp
         ca: self-sign
+      - name: certid
```
seocam, line 19: You need to verify this certificate in the task "Verify Certificate"

tests/tests_no_auto_renew.yml
```diff
@@ -24,6 +27,7 @@
         subject_alt_name:
           - DNS:www.example.com
         auto_renew: no
+      - path: /tmp/defaultcert
```
seocam, line 30: You need to add the `.crt` in the path. It would be also good to verify key-path and if `auto_renew: yes`

seocam:
> @amore17 @seocam are those test failures expected?

All Fedoras and CentOS/RHEL 8 should be passing.

PR #23, opened 2020-06-08
Added test case for Certmonger provider with IPA CA

seocam:


seocam:
Locally tests are passing but IPA installation is failing on CI. Still need to investigate the reason.
[ipa_tests_passing.log](https://github.com/linux-system-roles/certificate/files/4747851/ipa_tests_passing.log)

module_utils/certificate/providers/certmonger.py
```diff
@@ -53,6 +53,35 @@ def __init__(self, *args, **kwargs):
         self._certmonger_dbus = CertmongerDBus()
         self._certmonger_metadata = self._get_certmonger_request()
 
+    def module_input_validation(self):
+        """Validate module input."""
+        principal = self.module.params.get("principal")
```
richm, line 58: `get("principal", [])` - I think this needs to return an empty list so the `for` loop below won't fail - `[]` will still evaluate to boolean `False`

seocam: Fixed.

richm:
e.g. failure:
```
TASK [ipaserver : Install - Setup CA] ******************************************
task path: /tmp/freeipa-repo/roles/ipaserver/tasks/install.yml:251
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: RuntimeError: Certificate issuance failed (CA_UNREACHABLE: Error 7 connecting to http://ipaserver.test.local:8080/ca/ee/ca//profileSubmit: Couldn't connect to server.)
```
can we get the CA logs?

richm:
otherwise, lgtm

seocam:
> e.g. failure:
> 
> ```
> TASK [ipaserver : Install - Setup CA] ******************************************
> task path: /tmp/freeipa-repo/roles/ipaserver/tasks/install.yml:251
> An exception occurred during task execution. To see the full traceback, use -vvv. The error was: RuntimeError: Certificate issuance failed (CA_UNREACHABLE: Error 7 connecting to http://ipaserver.test.local:8080/ca/ee/ca//profileSubmit: Couldn't connect to server.)
> ```
> 
> can we get the CA logs?

IPA CA is a instance of dogtag pki (https://www.dogtagpki.org/wiki/PKI_Main_Page) which for some reason is not available in the local VM during installation. This has happened to be when I had issues with my DNS server. 

My plan was to actually merge #20 first and after look to this PR.

tests/tests_basic_ipa.yml
```diff
@@ -0,0 +1,44 @@
+---
+- name: Install IPA server
+  hosts: all
+  become: true
+  tasks:
+    - import_tasks: tasks/setup_ipa.yml
+
+- name: Issue simple self-signed certificate
```
nhosoi, line 8: self-signed ?

tests/tests_basic_ipa.yml
```diff
@@ -0,0 +1,44 @@
...
+  vars:
+    certificates:
+      - path: /etc/pki/tls/certs/mycert.crt
+        key_path: /etc/pki/tls/private/mycert.key
+        subject:
+          - O=TEST.LOCAL
+          - CN=ipaserver.test.local
+        subject_alt_name:
+          - DNS:ipaserver.test.local
+        # IPA CA doesn't respect the requested key_usage (and that's ok)
```
nhosoi, line 33: Is it something to be noted in README https://github.com/linux-system-roles/certificate/pull/30? (unless it's a common knowledge...)

seocam: Actually it is in the README. It's under section: https://github.com/linux-system-roles/certificate#certificate_requests
It's a generic note because it's pretty hard to know what are the exact fields a CA will or will not honor. I've added this extra comment in the tests so it doesn't look like we are performing a wrong verification. 
Do you think I should be more specific and maybe mention the README there?

nhosoi: Thank you, @seocam! I think this is clear enough.

t-woerner:
You are running into lots of OOM killer messages, please increase the memory for an IPA deployment to at least 2G

seocam:
[citest bad]

seocam:
[citest bad]

amore17:
On RHEL-7 tests failing for : Ensure python3 is installed

seocam:
[citest bad]

tests/tasks/setup_ipa.yml
```diff
@@ -2,7 +2,7 @@
 - name: Clone ansible-freeipa repo
   git:
     repo: https://github.com/freeipa/ansible-freeipa
-    version: v0.1.10
+    version: v0.1.12
```
nhosoi, line 5: I'm curious. In the future, if you update the certificate role and the latest ansible-freeipa version had been also upgraded then, you have to bump it manually to test against the latest like this? If it's not automated, could there be any chance to miss some unfortunate combinations? Or it's not an issue?

seocam: Correct @nhosoi. If freeipa project implement something new there might be updates in ansible-freeipa as well in order to get this feature available. In that case we might need to bump this version.

If we miss that the ipa tests will start to fail and we'd need to update that number. That could happen specially when there are major distro versions released (for example CentOS 8). 

I chose to use the fixed version instead of using git master because, I believe, the instabilities caused by eventual need for updates will be less frequent then the instabilities caused by in-development code.

At this time the update wasn't really required. I've just bumped version to leave it up-to-date.

tests/tasks/setup_ipa.yml
```diff
@@ -22,6 +22,12 @@
   hostname:
     name: ipaserver.test.local
 
+- name: Ensure nss package is up-to-date
+  package:
+    name: nss
+    state: latest
+  when: ansible_facts.distribution in ["CentOS", "RedHat", "Fedora"]
```
richm, line 29: `when: ansible_facts.os_family == "RedHat"`

richm:
[citest bad]

richm:
[citest bad]

richm:
[citest bad]

richm:
[citest]

richm:
so far, so good - with 4GB ram - I will continue to retest this throughout the day to see if I can get it to fail - @amore17 you said you had failures with 4GB - one difference is that you were using `m: 4072` which is less than 4GB `m: 4096` - I wouldn't think that would make a difference, but I don't know

richm:
[citest]

richm:
All tests passed - retry test to see if we can get a failure [citest]

richm:
All tests passed - retry test to see if we can get a failure [citest]

richm:
All tests passed - retry test to see if we can get a failure [citest]

richm:
All tests passed - retry test to see if we can get a failure [citest]

richm:
All tests passed - retry test to see if we can get a failure [citest]

richm:
All tests passed - retry test to see if we can get a failure [citest]

richm:
All tests passed - retry test to see if we can get a failure [citest]

richm:
All tests passed - retry test to see if we can get a failure [citest]

richm:
stopping the testing - it appears that there were no failures in many retries - 4GB seems ok

seocam:
@richm, @pcahyna, @nhosoi anything else to look/update here?

amore17:
Tested role of ansible-freeipa with rpm : https://github.com/linux-system-roles/certificate/pull/44

tests/tasks/setup_ipa.yml
```diff
@@ -1,11 +1,21 @@
 ---
+- name: Set __is_beaker_env
+  set_fact:
+    __is_beaker_env: "{{ lookup('env', 'BEAKERLIB') != '' }}"
+
+- name: Install ansible-freeipa
+  package:
+    name: ansible-freeipa
+  when: __is_beaker_env
```
richm, line 9: `delegate_to: localhost`

seocam: Updated

richm:
please add something like this https://github.com/richm/linux-system-roles-certificate/commit/61c4ec22c397c4200dfc1e95da7f19a72bb18bbd in order to capture the ipa logs when ipa installation fails

seocam:
> please add something like this [richm@61c4ec2](https://github.com/richm/linux-system-roles-certificate/commit/61c4ec22c397c4200dfc1e95da7f19a72bb18bbd) in order to capture the ipa logs when ipa installation fails

ok. I'm thinking about adding this inside the tasks file and perform the block/rescue in the ipaserver role inclusion. This way if we have more tests that requires IPA this will be also there. Does that make sense?

tests/tasks/setup_ipa.yml
```diff
@@ -17,17 +28,55 @@
...
+    name: nss
+    state: latest
+  when: ansible_facts.os_family == "RedHat"
+
+- name: Run IPA server install
+  block:
+    - include_role:
+        name: ipaserver
+    - fail:
+        msg: failed
```
richm, line 48: sorry - I left this in for debugging - please remove this fail task

richm: this needs to be moved to the end of the rescue block

seocam:
[citest bad]

seocam:
[citest bad]

richm:
[citest bad]

richm:
[citest bad]

richm:
[citest bad]

richm:
[citest]

PR #24, opened 2020-06-09
Added support to run scripts before/after cert is issued

seocam:


module_utils/certificate/providers/base.py
```diff
@@ -526,6 +528,29 @@ def certificate_key_path(self):
         """Path where key should be placed."""
         return self._get_store_location(key=True)
 
+    def _get_hooks_directory(self, dirname):
+        script_name = os.path.basename(self.module.params.get("name")) + ".sh"
+        provider_config_directory = self.module.params.get("provider_config_directory")
+        return os.path.join(provider_config_directory, dirname, script_name)
```
richm, line 534: do you need to do something like `mkdir -p $provider_config_directory/$dirname/$(dirname $script_name))`?

seocam: I've done that in the tasks. Do you think it's better to perform it here?

richm: > I've done that in the tasks. Do you think it's better to perform it here?

no, tasks is fine - I just didn't see it here in this file

richm: ok - I see where you create `$provider_config_directory/pre-scripts` and `post-scripts` - but e.g. if you have `name: /path/to/mycert` - don't you need to create `$provider_config_directory/pre-scripts/path/to` in order to write `$provider_config_directory/pre-scripts/path/to/mycert.sh`?

richm: ah, I see - there is a `basename` there

richm: There is a potential conflict - if the user specifies `name: /path/A/mycert` and `name: /path/B/mycert` - this code will attempt to create two different `pre-scripts/mycert.sh` - you might need to use the `sha` hash in the filename e.g. `pre-scripts/mycert-$hashval.sh` or something like that

seocam: well observed... it would be really nice if name was actually a single identifier but it's not...

tasks/main.yml
```diff
@@ -34,6 +34,34 @@
       when: |
         __certificate_provider_vars[__certificate_provider].packages is defined
 
+    - name: Ensure pre-scripts hooks directory exists
```
richm, line 37: The permissions and ownership of the script directories needs to be locked down.  I would suggest the certmonger user/group for the directory user/group, and `0700` for directory permission, and `0700` for file permission.

module_utils/certificate/providers/base.py
seocam: What about a method in the provider to define an ID for the CSR? Certmonger already have a request-id. This way we could suffix the script with that.

richm: not sure which way is better, but an alternative is to create `$provider_config_directory/pre-scripts/path/A/mycert.sh` and `$provider_config_directory/pre-scripts/path/B/mycert.sh`

richm: > What about a method in the provider to define an ID for the CSR? Certmonger already have a request-id. This way we could suffix the script with that.

ok - will you have the request-id at this point in the code?

seocam: We do. It's in `self._certmonger_metadata`.

richm: ok

tasks/main.yml
seocam: Ok. I'll create variables for that.

module_utils/certificate/providers/base.py
seocam: There is actually no certmonger used in centos/rhel. I'll leave group/owner as root and mode as 0770.

seocam: Amended.

PR #25, opened 2020-06-10
Test issuing certificate with certmonger provider.

amore17:


tests/tasks/setup_ipa.yml
```diff
@@ -0,0 +1,42 @@
...
+  loop:
+    - ipaserver
+    - ipaclient
+
+- name: Set hostname
+  hostname:
+    name: ipaserver.test.local
+
+- name: Check /etc/hosts
+  shell: cat /etc/hosts
```
nhosoi, line 26: We'd better use `command`? (the same is applied to tests/tests_basic_ipa.yml, as well.)
https://github.com/oasis-roles/meta_standards#ansible-best-practices

Plus, if it's just doing `cat /etc/hosts`, we could do this, too.
`   - debug: msg="{{ lookup('file', '/etc/hosts') }}"`

pcahyna: > We'd better use `command`? (the same is applied to tests/tests_basic_ipa.yml, as well.)
> https://github.com/oasis-roles/meta_standards#ansible-best-practices

yes

> 
> Plus, if it's just doing `cat /etc/hosts`, we could do this, too.
> ` - debug: msg="{{ lookup('file', '/etc/hosts') }}"`

No, because lookups are local (to the control host). If you want to use something more functional than `command`, use `slurp`. https://github.com/oasis-roles/meta_standards#check-mode-and-idempotency-issues

tests/tasks/setup_ipa.yml
```diff
@@ -0,0 +1,42 @@
+---
+- name: Clone ansible-freeipa repo
+  git:
+    repo: https://github.com/freeipa/ansible-freeipa
+    version: v0.1.10
+    dest: /tmp/freeipa-repo
+  delegate_to: 127.0.0.1
```
pcahyna, line 7: why not `localhost` ?

tests/tasks/setup_ipa.yml
```diff
@@ -0,0 +1,42 @@
+---
```
pcahyna, line 1: this file will conflict with #23 . Remove it from your commit and base your branch on top of #23.

tests/tasks/setup_ipa.yml
richm: since this is test code, I'm willing to allow the use of `command: cat /etc/hosts` rather than using `slurp` and having to base64 decode the output.  But for production code, `slurp` should be used.

amore17:
As this PR is only to debug ipa-server installation failure in CI.

PR #33, opened 2020-06-17
do not use jinja2 map filter with default param

richm:
support for the map filter with default was added in jinja2 2.11.0
https://jinja.palletsprojects.com/en/2.11.x/changelog/#version-2-11-0
But some of our supported platforms use jinja2 2.10, so we need to
rewrite the code to not use map/default.

Jinja2 gives warnings/errors like this:
```
TASK [linux-system-roles.certificate : Ensure provider packages are installed] ***
task path: tasks/main.yml:26
[WARNING]: Falling back to Ansible unique filter as Jinja2 one failed:
Unexpected keyword argument 'default'
...
TASK [linux-system-roles.certificate : Ensure certificate requests] ************
task path: tasks/main.yml:104
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: dbus.exceptions.DBusException: org.freedesktop.DBus.Error.ServiceUnknown: The name org.fedorahosted.certmonger was not provided by any .service files
failed: [image.qcow2] (item={'name': 'mycert', 'dns': 'www.example.com'....
```

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -16,12 +16,14 @@
 - name: Install and configure required providers
   vars:
     __certificate_providers: |
-      {{
-        certificate_requests |
-        map(attribute='provider', default=__certificate_provider_default) |
-        unique |
-        list
-      }}
+      {% set ns = namespace(unique_providers=[]) %}
+      {% for item in certificate_requests %}
+      {%   set _ = ns.unique_providers.append(
+             item.provider |
+             d(__certificate_provider_default)
+           ) %}
+      {% endfor %}
+      {{ ns.unique_providers | unique }}
   block:
     - name: Ensure provider packages are installed
       package:


seocam:
In this case, is it possible to use Jinja 2.10 in the tests? Is this Jinja executed in the controller? If not we could probable add jinja version here in the repo.

richm:
> In this case, is it possible to use Jinja 2.10 in the tests?

I'm trying to find out why some of our rhel8 tests use a jinja2 version which supports map/default and some do not.

> Is this Jinja executed in the controller?

Yes.

> If not we could probable add jinja version here in the repo.

For the CI tests, we use the jinja2 package provided by the OS, which is what we should be doing.

tasks/main.yml
```diff
@@ -16,12 +16,14 @@
...
     __certificate_providers: |
-      {{
-        certificate_requests |
-        map(attribute='provider', default=__certificate_provider_default) |
-        unique |
-        list
-      }}
+      {% set ns = namespace(unique_providers=[]) %}
+      {% for item in certificate_requests %}
+      {%   set _ = ns.unique_providers.append(
```
seocam, line 21: Why to we have to use `set _` here?

tasks/main.yml
```diff
@@ -16,12 +16,14 @@
 - name: Install and configure required providers
   vars:
     __certificate_providers: |
-      {{
-        certificate_requests |
-        map(attribute='provider', default=__certificate_provider_default) |
-        unique |
-        list
-      }}
+      {% set ns = namespace(unique_providers=[]) %}
```
seocam, line 19: I never used used `namespace`, why is it required?

richm: https://jinja.palletsprojects.com/en/2.10.x/templates/#assignments - see the entire "Scoping Behavior" section

tasks/main.yml
richm: not sure - I think in jinja2 you cannot have a "bare expression" e.g. something like
```python
for item in certificate_requests:
    ns.unique_providers.append(item.provider)
```
would be perfectly legal in python, but apparently jinja2 considers `ns.unique_providers.append(item.provider)` by itself to not be a valid statement, so you have to have the `set` with the "dummy" assignment to make it a real jinja2 statement.  We also ran into this with logging and solved it the same way.  If you can figure out a better way to do this, please do.

PR #34, opened 2020-06-29
RHELPLAN-46757 - bug with current implementation of platform/version …

nhosoi:
…specific vars/tasks file include/import

Updated based on the oasis-roles meta_standards:
  Supporting multiple distributions and versions
  - Always specify the explicit, absolute path to the files to be included
  - Using loop instead of with_first_found.
  - Renaming {RedHat,CentOS}-7.yml with {RedHat,CentOS}_7.yml, respectively.


tasks/main.yml
```diff
@@ -1,12 +1,15 @@
...
+- name: Set platform/version specific variables
   include_vars: "{{ item }}"
-  with_first_found:
-    - "{{ ansible_distribution }}-{{ ansible_distribution_version }}.yml"
-    - "{{ ansible_distribution }}-{{ ansible_distribution_major_version }}.yml"
-    - "{{ ansible_distribution }}.yml"
-    - "{{ ansible_os_family }}.yml"
-    - "default.yml"
+  loop:
+    - "{{ role_path }}/vars/default.yml"
```
pcahyna, line 5: This repeated `{{ role_path }}/vars` is not very aestehetical. Can you use the same or similar approach as in https://github.com/linux-system-roles/nbde_server/pull/9/files#diff-2444ad0870f91f17ca6c2a5e96b26823R8 ?

nhosoi: Hmmm, is it ok not to follow https://github.com/oasis-roles/meta_standards/#supporting-multiple-distributions-and-versions?

richm: This changes the "first found" semantics.  This change should not do that.  This change should:
* use role_path to use absolute path
* use the lookup function form instead of `with_first_found`
* use `ansible_facts` instead of `ansible_varname`

pcahyna: > This changes the "first found" semantics. This change should not do that. 

Ouch! I missed that.

richm: > Hmmm, is it ok not to follow https://github.com/oasis-roles/meta_standards/#supporting-multiple-distributions-and-versions?

Yes.  That is only for new code that can be designed so that you do not have to use "first found" semantics.  For existing code, to be safe, we have to preserve "first found" semantics.

nhosoi:
[citest]

nhosoi:
[citest pending]

pcahyna:
[citest commit:2025a369ef7d22827bfcd3d02f3852c1b5798abf]

PR #35, opened 2020-06-30
Add negative test-case for name and key_size.

amore17:


tests/tests_basic_self_signed.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: Issue simple self-signed certificate
-  hosts: all
+  hosts: localhost
```
seocam, line 3: Is that correct? I believe it should remain `all`.

seocam:
@richm, @pcahyna should we deal with invalid inputs (like chars in key size)? If so I believe the error msg should be a little better :)

richm:
> @richm, @pcahyna should we deal with invalid inputs (like chars in key size)? If so I believe the error msg should be a little better :)

In general we don't do input validation for system roles - there are just too many parameters, combinations, permutations, etc.  So, we pick and choose the most important ones.  If key_size is very important, then you can implement input validation for it.  But you will probably not be able to implement input validation for every single input parameter/key, or check for all invalid combinations/permutations thereof, so pick your battles wisely.

pcahyna:
@amore17 @seocam can you please remind us what the role does when given invalid inputs and what is the error message?
@richm concerning combinations, I think that a role should validate if the user provides an unsupported combination of otherwise supported parameters. Sometimes it is hard to tell for users what combinations are supported. (E.g. when some parameters are supported only by some providers and the user specifies a provider that does not support the provided parameters.) I also think that modules should verify the inputs (there are utility functions to do this for you, right?)
I agree that outside modules, the roles should not excessively validate provided parameters, an exception might be boolean normalization (`| bool`).

pcahyna:
> you will probably not be able to ... check for all invalid combinations/permutations thereof

I would say that the interface should be orthogonal enough to minimize the amount of invalid combinations of parameters. Of course, this may be easier said than done.

PR #36, opened 2020-07-06
Tests for idempotency check.

amore17:


tests/tests_repeatedly.yml
```diff
@@ -0,0 +1,90 @@
...
+  hosts: all
+  become: true
+
+  vars:
+    certificate_requests:
+      - name: mycert
+        dns: www.example.com
+        ca: self-sign
+  roles:
+    - linux-system-roles.certificate
```
richm, line 42: How does this test for idempotency?  That is, how are you verifying that the role didn't change anything?  I guess since `name` is a unique key, you are guaranteed that this did not add a new certificate request for `mycert`, but you have no way to know if Ansible reported `changed=True` for some task in the role. @pcahyna do you know if there is any way to do this?  AFAIK, Ansible does not expose the internal statistics gathering except upon the final output of the `ansible-playbook` command as `changed=N` where `N` is an integer.  As in this proposal to implement support for idempotency checking - https://github.com/linux-system-roles/timesync/pull/72/files#diff-cfddcd75655d3ce8647e77f40dd7baf9R48-R71 - which looks for `CHANGED=>` in the output

seocam:
Hello @amore17! Thanks for the PR.

I think we can use the actual tests to write idempotency checks instead of write new checks. I wrote an example here: https://github.com/linux-system-roles/certificate/compare/master...seocam:idempotence-example

Basically we include the role again (it will use the same variables) and then we verify if it hasn't changed anything.

seocam:
> Hello @amore17! Thanks for the PR.
> 
> I think we can use the actual tests to write idempotency checks instead of write new checks. I wrote an example here: [master...seocam:idempotence-example](https://github.com/linux-system-roles/certificate/compare/master...seocam:idempotence-example)
> 
> Basically we include the role again (it will use the same variables) and then we verify if it hasn't changed anything.

BTW, that's the idea but this example doesn't work. result.changed is always false no matter if the role changed something or not.

PR #41, opened 2020-07-09
Test issuing certificate with principal.

amore17:


tests/tests_principal.yml
```diff
@@ -0,0 +1,61 @@
...
+            oid: 1.3.6.1.5.2.2
+  tasks:
+    - name: Verify each certificate
+      include_tasks: tasks/assert_certificate_parameters.yml
+      loop: "{{ certificates }}"
+      loop_control:
+        loop_var: cert
+
+
+- name: Test issuing certificate with invalid principal.
```
seocam, line 40: I think I would use a different file for this test. Other then that it looks good.

amore17: Added changes

PR #44, opened 2020-07-14
Test with ipa as CA

amore17:


pcahyna:
Is it still Draft?
Is it intended to be merged eventually? #23 touched the same code.

seocam:
I think @amore17 was using this for test purposes only. Is that correct @amore17?

I suppose we can close.

amore17:
> I think @amore17 was using this for test purposes only. Is that correct @amore17?
> 
> I suppose we can close.

Yes closing this one.

PR #47, opened 2020-07-17
Using full imports instead of relative for module_utils

### LINKED ISSUE #46: With ansible-2.8 fails with No module named 'ansible.module_utils.certificate.providers.certmonger

seocam:
The bug was caused by relative imports inside module_utils. Certificate role have multiple files inside the "module_utils" directory and they are referenced by .something instead of ansible.module_utils.certificate.something. Using the full reference gets the problem fixed for ansible >=2.8.

* Closes #46

### DIFF
--- a/tests/tasks/assert_certificate_parameters.yml
+++ b/tests/tasks/assert_certificate_parameters.yml
@@ -12,7 +12,7 @@
   pip:
     name: certreader>=0.1.1
     virtualenv: "{{ __virtualenv_path }}"
-    virtualenv_command: python3 -m venv
+    virtualenv_command: /usr/bin/python3 -m venv
 
 - name: Retrieve certificate file stats
   stat:
--- a/module_utils/certificate/providers/__init__.py
+++ b/module_utils/certificate/providers/__init__.py
@@ -1,4 +1,6 @@
-from .certmonger import CertificateRequestCertmongerProvider
+from ansible.module_utils.certificate.providers.certmonger import (
+    CertificateRequestCertmongerProvider,
+)
 
 # fmt: off
 PROVIDERS = (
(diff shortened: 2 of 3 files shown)


tests/tasks/assert_certificate_parameters.yml
```diff
@@ -12,7 +12,7 @@
   pip:
     name: certreader>=0.1.1
     virtualenv: "{{ __virtualenv_path }}"
-    virtualenv_command: python3 -m venv
+    virtualenv_command: /usr/bin/python3 -m venv
```
richm, line 15: ?

seocam: This was required because my tests where failing with: 
`FAILED! => {"changed": false, "msg": "Failed to find required executable python3 -m venv in paths: /sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin"}`

I've added on same commit and add a note in the commit message. Do you think it's better to add an extra commit for that?

richm: I guess it is ok to leave it in, as it appears to be related to making certitificate work on ansible 2.8

But - do you have to specify `virtualenv_command`?  The `pip` module isn't smart enough to know to use `virtualenv` with python 2 and `python -mvenv` with python3?

richm: in ansible 2.8 the `pip` module expects `virtualenv_command` to be a single string: https://github.com/ansible/ansible/blob/stable-2.8/lib/ansible/modules/packaging/language/pip.py#L457 - so I'm not sure how it works with `/usr/bin/python3 -m venv`

richm: I see - it works by accident - because `os.path.basename("/usr/bin/python3 -m venv")` is `python3 -m venv` which is != `/usr/bin/python3 -m venv` . . .

PR #50, opened 2020-07-24
Add test for the include variable problem

pcahyna:
fixed in #34


pcahyna:
succeeding test: #51

pcahyna:
closing, I used it only to show the test failure

PR #55, opened 2020-08-07
Fix idempotency issue with earlier versions of certmonger.

### LINKED ISSUE #45: The role is not idempotent in CentOS/RedHat 7
Running any of `tests/tests_*.yml` more then once keeps returning `changed=1` in CentOS/RedHat 7

The expected result is that after the first run the next executions would have `changed=0`

Reproducible by: 
```
ansible-playbook --become -i <host_ip>, tests/tests_basic_self_signed.ym
```

Recap for first run:
```
PLAY RECAP ***********************************************************************************************
192.168.124.189            : ok=28   changed=5    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

Recap for second (or third...) run:
```
PLAY RECAP ***********************************************************************************************
192.168.124.189            : ok=28   changed=5    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

rjeffman:
In versions earlier than 0.79, certmonger does not respect the order of
keyusage values, and a new certificate would be issued even with the
same data. This fix sort the keyusage values when comparing an existing
certificate with CSR data, if an affected certmonger version is used,
ensuring idempotency.

Fixes #45

### DIFF
--- a/module_utils/certificate/providers/certmonger.py
+++ b/module_utils/certificate/providers/certmonger.py
@@ -145,6 +145,16 @@ class CertificateRequestCertmongerProvider(CertificateRequestBaseProvider):
         Since the provider base class doesn't know about CA this method
         needs to add the verification for CA change.
         """
+        # For earlier version of certmonger, keyusage order was not
+        # respected, so the keys need to be sorted to compare the
+        # certificates against the csr. If this is not done, a new
+        # certificate is issued, even if data has not changed leading
+        # to an idempotency issue.
+        require_keyusage_sort = self.certmonger_version < StrictVersion("0.79.0")
+        if require_keyusage_sort:
+            self.existing_certificate.cert_data.get("key_usage", []).sort()
+            self.csr.cert_data.get("key_usage", []).sort()
+
         needs_update = super(
             CertificateRequestCertmongerProvider, self
         ).cert_needs_update


seocam:
I've worked on this fix with @rjeffman so I'll skip the review at this time.

module_utils/certificate/providers/certmonger.py
```diff
@@ -145,6 +145,16 @@ def cert_needs_update(self):
         Since the provider base class doesn't know about CA this method
         needs to add the verification for CA change.
         """
+        # For earlier version of certmonger, keyusage order was not
+        # respected, so the keys need to be sorted to compare the
+        # certificates against the csr. If this is not done, a new
+        # certificate is issued, even if data has not changed leading
+        # to an idempotency issue.
+        require_keyusage_sort = self.certmonger_version < StrictVersion("0.79.0")
```
pcahyna, line 153: would it be a problem to omit the check and do the same thing always?

seocam: If we do that we would be ignoring the order that key usage is defined by the user and I suppose we don't want to do that for every case. Newer certmonger versions and other tools (like certbot) respect the other defined by the user and we are trying to do the same.

That's why we are checking certmonger version and only performing the work-around for old versions.

richm: This means you will be altering the key_usage field order on some platforms but not others.

Ideally, there would be a way to do a comparison of `self.existing_certificate.cert_data.key_usage` with `csr.cert_data.key_usage` without having to modify either of those in-place, altering the order that the user specified.  Some way to make a temporary sorted copy of both and compare the sorted copies, without having to change the actual real data in-place.

rjeffman: I think is is ok to do so. My question is, do we want to change the order specified by the user if we don't have to?

richm: > I think is is ok to do so. My question is, do we want to change the order specified by the user if we don't have to?

It is inconsistent.  On some platforms you will change the order and on some platforms you will not.

One way this might be problem is when the user upgrades the platform:
* run the certificate role with key_usage `[3, 2, 1]` -> stored as `[1, 2, 3]`
* upgrade the system
* run the certificate role with key_usage `[3, 2, 1]` -> a new cert will be issued because the order is preserved and now does not match `[1, 2, 3]`

Perhaps this is a rare corner case - but if you can figure out a way to compare the key_usage lists without having to change them in-place, you don't have to worry about this case or similar cases

seocam: I see your point @richm. We currently have the same inconsistency when updating the key size. If you change the key size on RHEL/CentOS 7 it will not be changed but when you update your system it will be.

If we decide to embrace the limitations of RHEL 7 we probably should revisit the decision regarding key size as well and update this code after RHEL/CentOS 7 is no longer supported by us.

seocam: > This means you will be altering the key_usage field order on some platforms but not others.
> 
> Ideally, there would be a way to do a comparison of `self.existing_certificate.cert_data.key_usage` with `csr.cert_data.key_usage` without having to modify either of those in-place, altering the order that the user specified. Some way to make a temporary sorted copy of both and compare the sorted copies, without having to change the actual real data in-place.

BTW, even if we change the order only during the comparison (instead of performing in place) when the user updates the system the order will be changed because in the new version of RHEL the order the user chose will be respected and in the previous version it was ignored.

Since we are updating the order in a version that the provider ignores the order it shouldn't make any difference to perform in place or in a memory copy.

richm: > > This means you will be altering the key_usage field order on some platforms but not others.
> > Ideally, there would be a way to do a comparison of `self.existing_certificate.cert_data.key_usage` with `csr.cert_data.key_usage` without having to modify either of those in-place, altering the order that the user specified. Some way to make a temporary sorted copy of both and compare the sorted copies, without having to change the actual real data in-place.
> 
> BTW, even if we change the order only during the comparison (instead of performing in place) when the user updates the system the order will be changed because in the new version of RHEL the order the user chose will be respected and in the previous version it was ignored.

I don't understand.  When we save the certificate request, do we save the order that the user specified?  If so, when you upgrade, why won't that order be respected?


> 
> Since we are updating the order in a version that the provider ignores the order it shouldn't make any difference to perform in place or in a memory copy.

I don't understand.

seocam: Let me try to exemplify the bug we are trying to fix:

Issuing certificate on CentOS 7:
1. The user uses the role to ensure a certificate with the KU: A, B, C
1. The role checks in certmonger and verify that the certificate doesn't exist yet.
1. The role requests the certificate to certmonger with the KU: A, B, C
1. Certmonger returns the certificate with the KU: C, A, B

Running again on CentOS 7:
1. The user uses the role to ensure a certificate with the KU: A, B, C
1. The role checks in certmonger and find a certificate with KU: C, A, B
1. The role requests the certificate to certmonger with the KU: A, B, C
1. Certmonger returns a new certificate with the KU: C, A, B

The scenario after the "fix":

Issuing certificate on CentOS 7:
1. The user uses the role to ensure a certificate with the KU: A, B, C
1. The role checks in certmonger and verify that the certificate doesn't exist yet.
1. The role requests the certificate to certmonger with the KU: A, B, C
1. Certmonger returns the certificate with the KU: C, A, B

Running again on CentOS 7:
1. The user uses the role to ensure a certificate with the KU: A, B, C
1. The role checks in certmonger and find a certificate with KU: C, A, B
1. The role considers that KU is the same in the request and certificate and doesn't issue a new certificate

Running again on CentOS 8:
1. The user uses the role to ensure a certificate with the KU: A, B, C
1. The role checks in certmonger and find a certificate with KU: C, A, B
1. The role requests the certificate to certmonger with the KU: A, B, C
1. Certmonger returns the certificate with the KU: A, B, C

richm: > **Let me try to exemplify the bug we are trying to fix:**
> 
> Issuing certificate on CentOS 7:
> 
>     1. The user uses the role to ensure a certificate with the KU: A, B, C
> 
>     2. The role checks in certmonger and verify that the certificate doesn't exist yet.
> 
>     3. The role requests the certificate to certmonger with the KU: A, B, C
> 
>     4. Certmonger returns the certificate with the KU: C, A, B
> 
> 
> Running again on CentOS 7:
> 
>     1. The user uses the role to ensure a certificate with the KU: A, B, C
> 
>     2. The role checks in certmonger and find a certificate with KU: C, A, B

What exactly does it mean to "check in certmonger"?  Does it check the actual certificate, or the certificate request?

> 
>     3. The role requests the certificate to certmonger with the KU: A, B, C
> 
>     4. Certmonger returns a new certificate with the KU: C, A, B
> 
> 
> **The scenario after the "fix":**
> 
> Issuing certificate on CentOS 7:
> 
>     1. The user uses the role to ensure a certificate with the KU: A, B, C
> 
>     2. The role checks in certmonger and verify that the certificate doesn't exist yet.
> 
>     3. The role requests the certificate to certmonger with the KU: A, B, C
> 
>     4. Certmonger returns the certificate with the KU: C, A, B
> 
> 
> Running again on CentOS 7:
> 
>     1. The user uses the role to ensure a certificate with the KU: A, B, C
> 
>     2. The role checks in certmonger and find a certificate with KU: C, A, B
> 
>     3. The role considers that KU is the same in the request and certificate and doesn't issue a new certificate
> 
> 
> Running again on CentOS 8:
> 
>     1. The user uses the role to ensure a certificate with the KU: A, B, C
> 
>     2. The role checks in certmonger and find a certificate with KU: C, A, B
> 
>     3. The role requests the certificate to certmonger with the KU: A, B, C
> 
>     4. Certmonger returns the certificate with the KU: A, B, C

seocam: It checks in certmonger metadata using dbus for it.

richm: > It checks in certmonger metadata using dbus for it.

And is that the cert itself, or the cert request?  I suppose it doesn't matter if certmonger doesn't preserve the order of the key_usage in the cert request?  This is what I thought was happening:

Request with no existing cert

1. certificate role generates a certificate request with key_usage `[C, A, B]`
2. certmonger stores the cert request
3. certmonger requests a cert with key_usage `[C, A, B]`
4. CA returns the cert with key_usage `[A, B, C]`
5. certificate role stores the cert with key_usage `[A, B, C]` in the cert store
6. certmonger stores the original certificate request with key_usage `[C, A, B]` in its cert request store

Request with existing cert

1. certificate role generates a certificate request with key_usage `[B, A, C]`
2. certmonger searches its cert request store to see if there is already a matching cert request - it finds one with key_usage `[C, A, B]`
3. the certificate role compares the stored certificate request with key_usage `[C, A, B]` with the new cert request with key_usage `[B, A, C]`

Is this not what's happening?  Is the certificate role using the cert from step 5 for the comparison?

seocam: @richm, today @rjeffman and I paired to debug this deeper and we found that the order is not kept even on CentOS/RHEL 8.
We'll close this PR for now and tomorrow we'll work to try to find the real issue (which is likely to be in our code Python rather then in certmonger).

PR #58, opened 2020-08-12
Fix idempotence issue with python version < 3.5

### LINKED ISSUE #45: The role is not idempotent in CentOS/RedHat 7

seocam:
For earlier versions of Python the dict order was not maintained and
comparing key_usage of a certificate and a new request resulted in a
new certificate request, even if the keys were not changed, only
reordered.

According to section 4.2.1.3 of RFC 5280 [1], the key_usage is a bit string
so, independent of the order the user requested, they will have the
same order in the certificate and certificate sign request.

Ordering the keys before comparing the request and the certificate
ensures idempotency, independent of python's version.

Fixes #45

[1] https://tools.ietf.org/html/rfc5280#page-30

### DIFF
--- a/module_utils/certificate/providers/base.py
+++ b/module_utils/certificate/providers/base.py
@@ -184,6 +184,9 @@ class CertificateProxy:
         if info.get("ip"):
             info["ip"] = [ipaddress.ip_address(to_text(ip)) for ip in info["ip"] if ip]
 
+        if info.get("key_usage"):
+            info["key_usage"] = set(info["key_usage"])
+
         if info.get("extended_key_usage"):
             info["extended_key_usage"] = [
                 cls._get_extended_key_usage_object_identifier(eku).dotted_string
@@ -317,7 +320,7 @@ class CertificateProxy:
     @property
     def key_usage(self):
         """Return the Key Usage in the certificate."""
-        return self.cert_data.get("key_usage") or []
+        return self.cert_data.get("key_usage") or set()
 
     @property
     def extended_key_usage(self):
@@ -363,18 +366,16 @@ class CertificateProxy:
         return san_values
 
     def _get_key_usage(self):
-        if not self._key_usage_ext:
-            return []
-
-        key_usages = []
-        for attr, key_usage in self.KEY_USAGE_ATTR_MAP.items():
-            try:
-                key_usage_enabled = getattr(self._key_usage_ext.value, attr)
-            except ValueError:
-                pass
-            else:
-                if key_usage_enabled:
-                    key_usages.append(key_usage)
+        key_usages = set()
+        if self._key_usage_ext:
+            for attr, key_usage in self.KEY_USAGE_ATTR_MAP.items():
+                try:
+                    key_usage_enabled = getattr(self._key_usage_ext.value, attr)
+                except ValueError:
+                    pass
+                else:
+                    if key_usage_enabled:
+                        key_usages.add(key_usage)
         return key_usages
 
     @classmethod


rjeffman:
I've worked on this fix with @seocam, so I won't review this PR.

module_utils/certificate/providers/base.py
```diff
@@ -430,6 +430,13 @@ def __eq__(self, other):
         self_info = {k: v for k, v in self.cert_data.items() if v}
         other_info = {k: v for k, v in other.cert_data.items() if v}
 
+        self._module.debug(
+            "Preparing CertificateProxy objects for comparison: "
+            "key_usage attribute will be sorted."
+        )
+        self_info.get("key_usage", []).sort()
+        other_info.get("key_usage", []).sort()
```
richm, line 438: This modifies the data in `self.cert_data['key_usage']` and in `other.cert_data['key_usage']` - that is, it modifies the data in-place as a side effect of the comparison:
```python
>>> self_cert_data = {'keyA':'valueA', 'key_usage':['C', 'B', 'A']}
>>> other_cert_data = {'keyA':'valueA', 'key_usage':['B', 'A', 'C']}
>>> self_info = {k: v for k, v in self_cert_data.items() if v}
>>> other_info = {k: v for k, v in other_cert_data.items() if v}
>>> self_info['key_usage'] is self_cert_data['key_usage']
True
>>> self_info.get("key_usage", []).sort()
>>> other_info.get("key_usage", []).sort()
>>> self_cert_data
{'keyA': 'valueA', 'key_usage': ['A', 'B', 'C']}
>>> other_cert_data
{'keyA': 'valueA', 'key_usage': ['A', 'B', 'C']}
```

pcahyna: it seems that a list is not the best data type to represent this information, could one use a  [set](https://docs.python.org/2/library/stdtypes.html#set-types-set-frozenset)?

seocam: That's true @richm. That side effect was unintended. We can use a copy of the key instead of just the using the key to avoid that.

@pcahyna using a sounds like a good idea and would probably work.

rjeffman: @pcahyna that's a great idea to use sets.

I modified the patch using sets for the key_usage parameter.

module_utils/certificate/providers/base.py
```diff
@@ -184,6 +184,9 @@ def load_from_params(cls, module, params):
         if info.get("ip"):
             info["ip"] = [ipaddress.ip_address(to_text(ip)) for ip in info["ip"] if ip]
 
+        if info.get("key_usage"):
+            info["key_usage"] = set(info["key_usage"])
```
richm, line 188: This doesn't preserve the order.  I think if you want to use a `set`, you should store the data as a `list`, to preserve the order, then create a temporary `set` variable copy of the list for comparison purposes only.

rjeffman: For the certificates, and the providers, the order is not important. It is only important when we were comparing lists in the role, to verify if a new certificate was needed to be issue or not.

I still don't get why we need to maintain the order.

seocam: @richm we don't need to preserve the order. We need to ensure that the order is ignored when we compare the objects. That's why @pcahyna suggested us to use a set. The order for the key_usage is actually irrelevant because it will be always stored in the same order.

richm: Ok, I get it.

What about the method `key_usage`?  https://github.com/linux-system-roles/certificate/blob/master/module_utils/certificate/providers/base.py#L317
should this return a `set()` instead of `[]` if no key usage was specified?

rjeffman: It should. I missed that. Fixed.

module_utils/certificate/providers/base.py
```diff
@@ -317,7 +320,7 @@ def email(self):
     @property
     def key_usage(self):
         """Return the Key Usage in the certificate."""
-        return self.cert_data.get("key_usage") or []
+        return set(self.cert_data.get("key_usage") or [])
```
richm, line 323: `self.cert_data.get("key_usage")` is already a `set` - and you can just call `set` with an empty constructor, you do not have to pass an empty `list` to the `set` constructor - I think this should be
`        return self.cert_data.get("key_usage") or set()`

rjeffman: Yes. I forgot _get_key_usage will always return a set.

PR #68, opened 2020-11-12
Use module_utils/certificate_lsr/ to avoid naming conflicts

richm:
module_utils is in the global namespace in Ansible.  The word
"certificate" is generic enough that someone else could create
a file or directory with this name.  Therefore, use
"certificate_lsr/" for the directory name to avoid any problems.
Note that the network and storage roles also do this.


rjeffman (approved):
lgtm

PR #72, opened 2020-12-23
use tox for ansible-lint instead of molecule

richm:
Deprecate the use of molecule for linting.  Use tox for ansible-lint.
The next step will be to remove the ansible-lint dependency from
the molecule environments.


richm:
please review
the next step after this PR will be to fix the centos6/centos7 molecule issues

rjeffman (approved):
LGTM.

PR #77, opened 2021-02-09
Use latest pip.

amore17:
Tests was failing with dependency issue while
installing certreader as pip was installed with older version.
Thus added fix to use latest pip.

Signed-off-by: Anuja More <amore@redhat.com>

### DIFF
--- a/tests/tasks/assert_certificate_parameters.yml
+++ b/tests/tasks/assert_certificate_parameters.yml
@@ -8,6 +8,13 @@
     name:
       - python3
 
+- name: Install the package, force upgrade
+  pip:
+    name: pip
+    state: latest
+    virtualenv: "{{ __virtualenv_path }}"
+    virtualenv_command: /usr/bin/python3 -m venv
+
 - name: Install certreader
   pip:
     name: certreader>=0.1.1


tests/tasks/assert_certificate_parameters.yml
```diff
@@ -8,6 +12,14 @@
     name:
       - python3
 
+- name: Install the package, force upgrade
+  pip:
+    name: pip
+    state: latest
+    virtualenv: "{{ __virtualenv_path }}"
+    virtualenv_command: /usr/bin/python3 -m venv
+  when: __is_beaker_env
```
rjeffman, line 21: Should we only update pip in Beaker? Or should we do it always? Are there any issues, regarding platform support, if pip is always updated?

If possible, I'd prefer that pip is always updated.

amore17: Agree I'll update it for all then.

richm: What is the issue that you are fixing by updating `pip`?

amore17: > What is the issue that you are fixing by updating `pip`?
In beaker environment tests was failing for installing certreader as in that environment pip was older
pip 9.0.3 from /usr/lib/python3.6/site-packages (python 3.6)

richm: > Should we only update pip in Beaker? Or should we do it always? Are there any issues, regarding platform support, if pip is always updated?
> 
> If possible, I'd prefer that pip is always updated.

If it is working now in upstream CI, then we don't have to do anything.

richm: . . . and it is not working in upstream CI - so we need this fix in both upstream CI and beaker CI

rjeffman: Here is the issue that sparked it all: https://github.com/pyca/cryptography/issues/5771

Another alternative, which I did not look into, might be to pin Python's cryptography to version 3.3.x (i.e.: < 3.4): https://github.com/pyca/cryptography/issues/5799

richm: > Another alternative, which I did not look into, might be to pin Python's cryptography to version 3.3.x (i.e.: < 3.4): pyca/cryptography#5799

Whatever you decide is fine with me.

rjeffman (approved):
lgtm.

PR #78, opened 2021-02-09
use molecule v3, drop v2 - use tox-lsr 2.1.2

richm:
Switch to molecule v3.  This drops support for v2.
This requires tox-lsr 2 or later.
Upgrade to tox-lsr 2.1.2 to pick up fix for collection dependency
on python six.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -3,7 +3,7 @@ name: tox
 on:  # yamllint disable-line rule:truthy
   - pull_request
 env:
-  TOX_LSR: "git+https://github.com/linux-system-roles/tox-lsr@1.0.2"
+  TOX_LSR: "git+https://github.com/linux-system-roles/tox-lsr@2.1.2"
   LSR_ANSIBLES: 'ansible==2.8.* ansible==2.9.*'
   LSR_MSCENARIOS: default
   LSR_EXTRA_PACKAGES: libdbus-1-dev
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -3,9 +3,7 @@
 dependency:
   name: galaxy
 driver:
-  name: docker
-lint:
-  enabled: false
+  name: ${LSR_MOLECULE_DRIVER:-docker}
 platforms:
   - name: centos-7
     image: registry.centos.org/centos/systemd:latest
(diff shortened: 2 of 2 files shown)


richm:
hmm - `black` doesn't like the code converted to collection format because the line length is too long and not wrapped properly - @nhosoi what do you think - should we skip the `black` test for python code converted to collection format?  I'm not sure how to suppress errors like this, and not suppress other sorts of errors

rjeffman:
I suggest using "# fmt: off"  / "# fmt: on" around the imports and keeping black.

I'd also suggest the "()"-style import.

rjeffman (approved):
lgtm.

nhosoi:
> I suggest using "# fmt: off" / "# fmt: on" around the imports and keeping black.
> 
> I'd also suggest the "()"-style import.

+1

FYI: we put `  # noqa:E501` in the network tests.

richm:
> I'd also suggest the "()"-style import.

Can we do that even if the line is short and doesn't need wrapping?  Will that cause any problems with `black`, `pylint`, or `flake8`?

rjeffman:
> > I'd also suggest the "()"-style import.
> 
> Can we do that even if the line is short and doesn't need wrapping? Will that cause any problems with `black`, `pylint`, or `flake8`?

Unfortunately, no. `pylint` and `flake8` don't care much about it, but `black` complains if the code could be written in a single line. I might have been thinking about ansible-freeipa, where I want to change our (loong list of) imports to a black-friendly style.

I think the best solution is to use `#fmt: off` and `# noqa: E501` on imports that will be modified for collections.

richm:
> > > I'd also suggest the "()"-style import.
> > 
> > 
> > Can we do that even if the line is short and doesn't need wrapping? Will that cause any problems with `black`, `pylint`, or `flake8`?
> 
> Unfortunately, no. `pylint` and `flake8` don't care much about it, but `black` complains if the code could be written in a single line.

Right - so we can't modify the original source to do this in such a way that black will pass in both the original source and the collection converted source :-(

>I might have been thinking about ansible-freeipa, where I want to change our (loong list of) imports to a black-friendly style.
> 
> I think the best solution is to use `#fmt: off` and `# noqa: E501` on imports that will be modified for collections.

The problem is that we don't necessarily know ahead of time which imports will need to be modified for collections, and I don't want to modify every single `import` in the module code, and it would be very difficult for the collection conversion script to insert such directives into the converted code.

PR #84, opened 2021-03-12
RHELPLAN-69928  Add a note to each module Doc to indicate it is private

nhosoi:
Adding "WARNING: Do not use this module directly! It is only for role internal use." to certificate_request DOCUMENTATION.


richm (approved):
lgtm

PR #90, opened 2021-07-20
Instead of the archive module, use "tar" command for backup.

nhosoi:
Note: having the module 'archive' makes the role fail with an error
"couldn't resolve module/action 'archive'." if executed with ansible-
navigator.

Ref: bz1984182


richm (approved):
lgtm

nhosoi:
Thank you for your reviews, @richm.

@rjeffman, could you please take a look? And if it looks ok, may I ask you to merge since I have no privilege for the action. Thanks.

rjeffman (approved):
LGTM.

Note that setting pipefail wil fail if we test in Ubuntu/Debian which uses `dash` instead of `bash`.

PR #92, opened 2021-08-06
Instead of the unarchive module, use "tar" command for backup.

nhosoi:
Also, fixing a warning issue: Do not warn about unarchive.

Ref: bz1984182, bz1987096


richm (approved):
lgtm

PR #94, opened 2021-10-07
support ansible-core 2.11 ansible-test and ansible-lint

richm:

PR #95, opened 2021-10-19
tests: Install latest certmonger from updates-testing

martinpitt:
Current Fedora certmonger has a regression with parsing `Extension::`
fields (https://pagure.io/certmonger/issue/223). This is fixed in
updates-proposed. Ensure the tests install that package to unbreak them.

----

This should fix at least the Fedora failures in [recent PRs](https://github.com/linux-system-roles/certificate/pull/94). This is obviously a hack, I don't know how often your test images get refreshed. It's also fine for me to not merge this and just wait for the fix to land, but it somehow breaks *all* operating systems (including CentOS/RHEL 7). Something recent from Fedora is clearly leaking into these VMs, but I didn't track down what exactly. This should help at least a little bit, so that I can start sending some actual fixes.

### DIFF
--- a/tests/tasks/setup.yml
+++ b/tests/tasks/setup.yml
@@ -1,0 +1,7 @@
+# These steps run for *all* tests
+
+- name: setup - get fix for https://pagure.io/certmonger/issue/223
+  dnf:
+    name: certmonger
+    state: latest
+    enablerepo: updates-testing
--- a/tests/tests_basic_self_signed.yml
+++ b/tests/tests_basic_self_signed.yml
@@ -1,4 +1,9 @@
 ---
+- name: Setup
+  hosts: all
+  tasks:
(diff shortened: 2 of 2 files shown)


rjeffman:
So, the reason why certificate roles are failing is due to the way the certificates are validated.

In `tests/tests/assert_certificat_parameters.yml`a virtual environment is created, and certreader is installed via pip. This brings `cryptography == 35.0.0`, which is the version that change the certificate parser to a Rust based one that is more strict than the older. A temporary workaround for this issue will be available in a newer version, and this is the version that will be supported on RHEL, so we have some time to fix everything, or have certificates renewed by a patched certmonger.

martinpitt:
Thanks @rjeffman ! This makes so much more sense!

PR #96, opened 2021-10-19
Fix parser fail on certificate verification.

rjeffman:
Due to a change in Python's cryptography version 35.0.0 certificate
parser, and a difference in the ASN.1 certificate spec interpretation,
the certificates generated by certmonger fail to be validated.

This patch forces the version for the 'cryptography' package installed
to ignore the affected version, and should allow the tests for this
role to be executed.

certmonger already has a fix for the issue, but it might not be
available for every release supported by certificate role.


richm:
[citest]

PR #97, opened 2021-10-20
Fix certificate permissions with "group" option

martinpitt:
The default permissions are 0600, so that certificates are inaccessible
to the specified group. Add group read permission in that case, so that
this actually works.

Test this for both local certmonger and IPA.

-----

I tested this locally against a Fedora 34 cloud image.


rjeffman:
Failed and pending (which actually failed) tests seems to fail due to infra issues. I'm ok on merging this PR.

@richm what do you think?

PR #106, opened 2022-02-01
Add certificate profile

ooraini:
Notes:
Changing the profile will not delete and re-issue the certificate.
I'm not sure how would I write a test with the self-signed provider. I have tested it with IPA and it works as expected.

PR #107, opened 2022-02-01
fix python black errors

richm:


richm:
fix the errors seen in https://github.com/linux-system-roles/certificate/pull/106

PR #115, opened 2022-02-21
Remove the unnecessary code

nhosoi:
In the condition of param_sha1 == file_sha1, the condition "param !=
cur_script (contents of file)" never be true and never be executed.
It was introduced in commit e5fe0392f5491293719eddbf5afb29611748c55c
System Roles should consistently use ansible_managed in configuration files it manages

Replace ansible_managed_new with __header

@richm, @rjeffman, so sorry... I revisited the commit e5fe0392f5491293719eddbf5afb29611748c55c and realized it introduced the code which never been executed. This patch removes it. It does not do a bad thing, but we should not include the garbage... :(

And if ok, I'd like to replace the parameter `ansible_managed_new` with `__header` to adjust to the other roles. Thanks.

### DIFF
--- a/library/certificate_request.py
+++ b/library/certificate_request.py
@@ -160,7 +160,7 @@ options:
     description:
       - Command that should run after saving the certificate.
     required: false
-  ansible_managed_new:
+  __header:
     description:
       - Ansible ansible_managed string to put in header of file
       - should be in the format of {{ ansible_managed | comment }}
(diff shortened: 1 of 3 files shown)


richm (approved):
lgtm - note that it may take a few hours to get CI test results - all of the runners are very busy

rjeffman (approved):
lgtm. the new variable name is much better!

PR #116, opened 2022-03-11
Let each test use a different certificate file name

nhosoi:
This is part of the effort to allow CI tests run in the serialized
manner on one VM for shortening the duration of the CI tests. In
the current tests, the same certificate file path is shared and
there is no cleaning up implemented. When multiple tests run
sequentially, some attribute, e.g., the file ownership, could be
inherited from the previous tests, which could make the test fail
in the attribute verification. To avoid the false negative failure,
this commit chooses different certificate file name per test.

### DIFF
--- a/tests/tests_basic_ipa.yml
+++ b/tests/tests_basic_ipa.yml
@@ -11,7 +11,7 @@
 
   vars:
     certificate_requests:
-      - name: mycert
+      - name: mycert_basic_ipa
         dns: ipaserver.test.local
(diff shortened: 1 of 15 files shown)


spetrosi (approved):
lgtm

richm (approved):
lgtm

richm:
@rjeffman any objections?

rjeffman (approved):
lgtm

rjeffman:
Sorry for the late response. Last days of PTO. ;-)

PR #130, opened 2022-07-29
Move Debian to Python 3 packages

martinpitt:
The python 2 packages don't exist any more in current stable Debian 11
and Ubuntu 22.04 LTS. Use the python3-* packages (vars/main.yml has the
correct ones).

---

Spotted in https://github.com/linux-system-roles/cockpit/pull/68 , this role does not work at all any more in current Debian/Ubuntu.

### DIFF
--- a/vars/Debian.yml
+++ b/vars/Debian.yml
@@ -1,10 +1,5 @@
 # SPDX-License-Identifier: MIT
 ---
-# Put internal variables here with Debian 10 specific values.
-
-__certificate_packages:
-  - python-pyasn1
-  - python-cryptography
-  - python-dbus
+# Put internal variables here with Debian 11 specific values.
 
 __certificate_default_directory: /etc/ssl


richm:
I'll have to defer to @rjeffman about Debian support

martinpitt:
@rjeffman Friendly review ping? Thanks!

vars/Debian.yml
```diff
@@ -1,10 +1,5 @@
 # SPDX-License-Identifier: MIT
 ---
-# Put internal variables here with Debian 10 specific values.
-
-__certificate_packages:
-  - python-pyasn1
-  - python-cryptography
-  - python-dbus
```
martinpitt, line 8: FTR: The python3 packages have existed for ages, at least since stretch. [cryptography](https://packages.debian.org/search?keywords=python3-cryptography), [pyasn1](https://packages.debian.org/search?keywords=python3-pyasn1), [dbus](https://packages.debian.org/search?keywords=python3-dbus)

rjeffman:
Hi @martinpitt, I'm getting back from PTO today, and will look at this soon (please ping me if I don't answer during this week).

On ansible-freeipa we've had some issues with Debian/Ubuntu support regarding Python 2/3 changes, due to packaging on Debian side. I'd like to have support for Ubuntu 20.04+ if Debian/Ubuntu is to be supported, and if at all possible.

Another issue is that freeipa server is currently required (as it is the only available provider, IIRC), but the server is not available in any recent Ubuntu version (last I know of is 18.04).

Anyway, I'll get back to you later this week.

martinpitt:
@rjeffman : FreeIPA is available [in all Ubuntu releases](https://launchpad.net/ubuntu/+source/freeipa). You probably meant that it is not available in some Debians? The [tracker](https://tracker.debian.org/pkg/freeipa) shows that it is not present in Debian 11 (latest stable release), but at least it's in backports. It tends to fall out of testing as it keeps getting release-critical bugs.

> I'd like to have support for Ubuntu 20.04+ if Debian/Ubuntu is to be supported, and if at all possible.

Right, me too. And the way to do that is to use the Python 3 packages, they have worked since oldstable and Ubuntu 18.04 at least.

rjeffman (approved):
LGTM.

rjeffman:
FTR: Tested with Debian 11, and it also worked.

PR #132, opened 2022-08-18
WIP: Use distro python cryptography

rjeffman:
Installing Python's cryptography from pip might require that a Rust
compiler is available on the distribution and platform. By using the
distro's package manager to install it, all dependencies will be
available on every supported platform.

### DIFF
--- a/tests/tasks/assert_certificate_parameters.yml
+++ b/tests/tasks/assert_certificate_parameters.yml
@@ -18,7 +18,7 @@
 - name: Ensure Python's cryptography is installed
   package:
     name: python3-cryptography
-    status: present
+    state: present
 
 - name: Install certreader
   pip:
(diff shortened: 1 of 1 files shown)


tests/tasks/assert_certificate_parameters.yml
```diff
@@ -15,10 +15,14 @@
     virtualenv: "{{ __virtualenv_path }}"
     virtualenv_command: /usr/bin/python3 -m venv
 
+- name: Ensure Python's criptography is installed
```
richm, line 18: cryptography

richm:
[citest]

richm:
[citest]

richm:
[citest]

richm:
Is this ready to be merged?

rjeffman:
Sorry, I forgot to set it as draft.

This will not fix the issue we were having on non-x86 platforms, as certreader needs to be replaced.

I'll work again on this in early Q4.

rjeffman:
This PR is not needed anymore. The actual work for the fix is PR #180

PR #134, opened 2022-08-21
Setting the group parameter changes file permissions

jh23453:
https://github.com/linux-system-roles/certificate/pull/97 has added the functionality. Give a hint in the README how it will work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Alternative Names (SAN).
 | ip                   | IP, or list of IPs, to be included in the certificate. IPs can be IPv4, IPv6 or both. Also can provide the default value for [common\_name](#common_name). | str or list | no | - |
 | auto_renew           | Indicates if the certificate should be renewed automatically before it expires.                   | bool        | no       | yes                     |
 | owner                | User name (or user id) for the certificate and key files.                                         | str         | no       | *User running Ansible*  |
-| group                | Group name (or group id) for the certificate and key files.                                       | str         | no       | *Group running Ansible* |
+| group                | Group name (or group id) for the certificate and key files. If a group is specified the resulting key/cert files are readable by the group. | str         | no       | *Group running Ansible* |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,7 +56,7 @@ Alternative Names (SAN).
 | ip                   | IP, or list of IPs, to be included in the certificate. IPs can be IPv4, IPv6 or both. Also can provide the default value for [common\_name](#common_name). | str or list | no | - |
 | auto_renew           | Indicates if the certificate should be renewed automatically before it expires.                   | bool        | no       | yes                     |
 | owner                | User name (or user id) for the certificate and key files.                                         | str         | no       | *User running Ansible*  |
-| group                | Group name (or group id) for the certificate and key files.                                       | str         | no       | *Group running Ansible* |
+| group                | Group name (or group id) for the certificate and key files. If a group is specified the resulting key/cert files are readable by the group. | str         | no       | *Group running Ansible* |
```
martinpitt, line 59: I find this redundant. "group for the cert/key files" is already meant to say exactly that. If it's not understandable, let's condense this into one sentence. Like "Additional group name or id which can read the certificate and key files"?

rjeffman (commented):
I wonder why `owner` is clear enough and `group` is not.

I'd like to keep both parameters with similar description.

README.md
jh23453: Let's see where I came from. I used the role to create the certificate without specifying owner and group. The permissions were 600 for the owner.
I used ipa-getcert in the past an specified both --cert-owner and --cert-perms to make the certificate group readable (need to check if that's really needed).
So, I looked to transfer the permission setting into the ansible role and missed that the group automatically has read permissions and looked for the (unneeded) parameter to change the permissions.
If you still think this pull request is not needed - that's fine with me.

richm:
Do we still need this PR?  If so, please rebase and fix the conflicts in README.md

richm:
closing due to inactivity

PR #142, opened 2022-12-13
Add check for non-inclusive language

nhosoi:
Add a check for usage of terms and language that is considered
non-inclusive. We are using the woke tool for this with a wordlist
that can be found at
https://github.com/linux-system-roles/tox-lsr/blob/main/src/tox_lsr/config_files/woke.yml

Note: this commit uses the customized woke placed locally in
.github/actions/custom-woke-action. It will be replaced with
the official woke once get-woke/woke#252
(Add an option "--count-only-error-for-failure") is processed.
    
CHANGELOG.md - cleanup non-inclusive words.
tests/tasks/setup_ipa.yml - Apply "wokeignore:rule"

### DIFF
--- a/.github/workflows/woke.yml
+++ b/.github/workflows/woke.yml
@@ -1,0 +1,18 @@
+# yamllint disable rule:line-length
+name: Check for non-inclusive language
+on:  # yamllint disable-line rule:truthy
+  - pull_request
+jobs:
+  woke:
+    name: woke
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: woke
+        uses: get-woke/woke-action@v0
+        with:
+          woke-args: "-c https://raw.githubusercontent.com/linux-system-roles/tox-lsr/main/src/tox_lsr/config_files/woke.yml"
+          # Cause the check to fail on any broke rules
+          fail-on-error: true
--- a/tests/tasks/setup_ipa.yml
+++ b/tests/tasks/setup_ipa.yml
@@ -10,9 +10,9 @@
   when: __is_beaker_env
 
 - name: Clone ansible-freeipa repo
+  # noqa git-latest
   git:
     repo: https://github.com/freeipa/ansible-freeipa
-    version: master
     dest: /tmp/freeipa-repo
   delegate_to: 127.0.0.1
   become: false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ correct ones).
 
 ### Other Changes
 
-- changelog_to_tag action - support other than "master" for the main branch name (#129)
+- changelog_to_tag action - github action ansible test improvements
 
 [1.1.6] - 2022-07-19
 --------------------
@@ -183,14 +183,14 @@ must be a `string` value like `"2.9"`, not a `float` value like `2.9`.
 
 ### Bug fixes
 
-- Fix some ansible-test sanity errors; suppressing the rest
+- Fix some ansible-test errors; suppressing the rest
 - Fix ansible-test errors
 - Add a note to each module Doc to indicate it is private
 
 ### Other Changes
 
 - Remove python-26 environment from tox testing
-- update to tox-lsr 2.4.0 - add support for ansible-test sanity with docker
+- update to tox-lsr 2.4.0 - add support for ansible-test with docker
 - CI: Add support for RHEL-9
 
 [1.0.2] - 2021-02-12


nhosoi:
[citest]

.github/workflows/woke.yml
```diff
@@ -0,0 +1,18 @@
...
+# name: woke
+on:  # yamllint disable-line rule:truthy
+  - pull_request
+jobs:
+  woke:
+    name: woke
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
richm, line 11:
```suggestion
        uses: actions/checkout@v3
```

nhosoi:
[citest]

tests/tasks/setup_ipa.yml
```diff
@@ -12,6 +12,7 @@
 - name: Clone ansible-freeipa repo
   git:
     repo: https://github.com/freeipa/ansible-freeipa
+    # wokeignore:rule=master
     version: master
```
richm, line 16: Here you can omit `version: master` - by default it will grab whatever branch is configured to be the default branch

nhosoi:
> So if you omit the `version: master` from the setup_ipa, and edit the CHANGELOG.md to remove references to `sanity`, can you use the official woke github action instead of the custom one?

Yes, we can. Do we want to avoid using `ansible-test sanity`? Just say `ansible-test`?

richm:
> > So if you omit the `version: master` from the setup_ipa, and edit the CHANGELOG.md to remove references to `sanity`, can you use the official woke github action instead of the custom one?
> 
> Yes, we can. Do we want to avoid using `ansible-test sanity`? Just say `ansible-test`?

Yes, just say `ansible-test`

tests/tasks/setup_ipa.yml
nhosoi: ansible-lint reports this error if no `version: master` is specified.
```
Error: git-latest Git checkouts must contain explicit version
You can skip specific rules or tags by adding them to your configuration file:
# .ansible-lint
warn_list:  # or 'skip_list' to silence them completely
  - git-latest  # Git checkouts must contain explicit version

Finished with 1 failure(s), 0 warning(s) on 82 files.
ERROR: InvocationError for command /home/runner/work/certificate/certificate/.tox/ansible-lint/bin/ansible-lint -v --exclude=tests/roles -c .ansible-lint (exited with code 2)
```
Which do we prefer between having `version: master` preceded by `# wokeignore:rule=master` or adding `- '403'` to the skip-list in `.ansible-lint`?

nhosoi: This commit https://github.com/linux-system-roles/certificate/pull/142/commits/604afea5eb3c8dc204709816745c95c31ea80da2 takes the latter.

richm: I would prefer noqa latest - https://ansible-lint.readthedocs.io/rules/latest/
```
  # noqa latest
  git:
    repo: https://github.com/freeipa/ansible-freeipa
    # wokeignore:rule=master
    version: master
```
I much, much prefer putting such suppressions inline in the code rather than the config file
* the suppression is next to the problematic code
* easier to deal with conversion to collection - don't have to merge a bunch of .ansible-lint files together
* doesn't affect all files

(/me wishes the error message said this rather than "just add it to the .ansible-lint file" :-(

nhosoi: Hmmm., this is the diff I made.
```
@@ -10,6 +10,7 @@
   when: __is_beaker_env
 
 - name: Clone ansible-freeipa repo
+  # noqa latest
   git:
     repo: https://github.com/freeipa/ansible-freeipa
     dest: /tmp/freeipa-repo
```
And I ran `tox -e ansible-lint`, which reported:
```
git-latest: Git checkouts must contain explicit version
tests/tasks/setup_ipa.yml:12 Task/Handler: Clone ansible-freeipa reported

You can skip specific rules or tags by adding them to your configuration file:
# .ansible-lint
warn_list:  # or 'skip_list' to silence them completely
  - git-latest  # Git checkouts must contain explicit version

Finished with 1 failure(s), 0 warning(s) on 86 files.
ERROR: InvocationError for command /home/nhosoi/linux-system-roles/certificate/.tox/ansible-lint/bin/ansible-lint -v --exclude=tests/roles --exclude=.github --exclude=.yamllint.yml -c .ansible-lint (exited with code 2)
```
I wonder `# noqa` works in the yaml file?

tests/tasks/setup_ipa.yml
```diff
@@ -53,6 +53,7 @@
   rescue:
     - name: FAILURE - check entropy
       command: cat /proc/sys/kernel/random/entropy_avail
+      changed_when: false
```
richm, line 56: don't worry about these - we'll change these later when we make all of the roles work with ansible-lint 6.x - looks like we have some ansible-core 2.14 work to do as well - `warn: false` breaks with 2.14

rafasgj:
About the 'version: master' on ansible-freeipa checkout, since I lost the argument to change it to "main", as we use it only to deploy IPA, and there won't be many changes, specially on what we need, we could pin the version to 'v1.9.0", and avoid the use of "master".

tests/tasks/setup_ipa.yml
```diff
@@ -10,9 +10,9 @@
   when: __is_beaker_env
 
 - name: Clone ansible-freeipa repo
+  # noqa git-latest
   git:
     repo: https://github.com/freeipa/ansible-freeipa
-    version: master
```
rafasgj, line 15: Here's what I meant about the use of 'master'. We can safely pin version 'v1.9.0' and update it if we need in the future.

On the maintainance cost of updating it, I believe it will be updated soon, but not often.

richm: Why can't we just omit `version` altogether, then git will use whatever is the default branch?

rafasgj: If we can, it would be better. I though it was required.

nhosoi: It is not required by the `git module`, but it is by `ansible-lint`. So, we are adding this line to skip the check by `ansible-lint`.
https://github.com/linux-system-roles/certificate/pull/142/files#diff-d1ca26ec3f441ccd0c4bb18b601d19668a8ffacfab0c71694fea7b46adc9823cR13

nhosoi:
@richm, @rafasgj, @spetrosi, could you please review this pr one more time? Thanks!

rjeffman (approved):
@nhosoi, although I don't understand why we can't pin the ansible freeipa version (and then don't have the `# noqa` comment, this change looks good to me.

And the use of woke looks promising, specially for the future. Thank you for looking into this issue! 

I'm approving these changes.

richm:
> @nhosoi, although I don't understand why we can't pin the ansible freeipa version (and then don't have the `# noqa` comment, this change looks good to me.

My only concern is that it stays on an old version for a long time and never gets updated, which is what has happened in the past:
https://github.com/linux-system-roles/certificate/commit/e84385a369969af5bdb64481934f6abb3204012a
https://github.com/linux-system-roles/certificate/commit/cc7d5a953bce80151c037decbf952e3993d56721#diff-d1ca26ec3f441ccd0c4bb18b601d19668a8ffacfab0c71694fea7b46adc9823cL15

So if we have some (preferably automated) process for regularly updating the version, that's fine with me.

> 
> And the use of woke looks promising, specially for the future. Thank you for looking into this issue!
> 
> I'm approving these changes.

nhosoi:
> My only concern is that it stays on an old version for a long time and never gets updated, which is what has happened in the past: [e84385a](https://github.com/linux-system-roles/certificate/commit/e84385a369969af5bdb64481934f6abb3204012a) [cc7d5a9#diff-d1ca26ec3f441ccd0c4bb18b601d19668a8ffacfab0c71694fea7b46adc9823cL15](https://github.com/linux-system-roles/certificate/commit/cc7d5a953bce80151c037decbf952e3993d56721#diff-d1ca26ec3f441ccd0c4bb18b601d19668a8ffacfab0c71694fea7b46adc9823cL15)
> 
> So if we have some (preferably automated) process for regularly updating the version, that's fine with me.

Me, too. @rjeffman, do you want me to revert the `version: master` line? Or any specific version in your mind? And maybe, we'd like to have a tiny github action to check the `ansible-freeipa` version?

rjeffman:
> Me, too. @rjeffman, do you want me to revert the `version: master` line? Or any specific version in your mind? And maybe, we'd like to have a tiny github action to check the `ansible-freeipa` version?

Hmm... I have bug to fix where an external dependency is broken. Maybe we could have a Github action that setup all the testing dependencies.

But that's for the future, I'm fine with this PR as it is.

richm:
[citest]

PR #157, opened 2023-02-02
Fix assert in tests_run_hooks.yml

nhosoi:
In the tests_run_hooks.yml test script, allow the timestamp of run_before and run_after to be equal to the certificate timestamp.

### DIFF
--- a/tests/tests_run_hooks.yml
+++ b/tests/tests_run_hooks.yml
@@ -51,16 +51,16 @@
         - name: Assert file created before cert
           assert:
             that:
-              - before_result.stat.mtime < cert_result.stat.mtime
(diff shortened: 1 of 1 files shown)


nhosoi:
[citest]

richm:
I verified the fix works on aarch64

nhosoi:
> I verified the fix works on aarch64

Thank you, @richm ! I was wondering how I could get standard-test-roles-inventory-qemu... :p

If it's not hard to run a test, could you please try this pr, as well?
https://github.com/linux-system-roles/ha_cluster/pull/106
Failed tests are tests_cib_properties_empty.yml and tests_cib_properties_one_set.yml.

PR #169, opened 2023-04-30
docs: Consistent contributing.md for all roles - allow role specific contributing.md section

richm:
Provide a single, consistent contributing.md for all roles.  This mostly links to
and summarizes https://linux-system-roles.github.io/contribute.html

Allow for a role specific section which typically has information about
role particulars, role debugging tips, etc.

See https://github.com/linux-system-roles/.github/pull/19

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -1,0 +1,54 @@
+Contributing to the certificate Linux System Role
+=================================================
+
+Where to start
+--------------
+
+The first place to go is [Contribute](https://linux-system-roles.github.io/contribute.html).
+This has all of the common information that all role developers need:
+
+* Role structure and layout
+* Development tools - How to run tests and checks
+* Ansible recommended practices
+* Basic git and github information
+* How to create git commits and submit pull requests
+
+**Bugs and needed implementations** are listed on
+[Github Issues](https://github.com/linux-system-roles/certificate/issues).
+Issues labeled with
+[**help wanted**](https://github.com/linux-system-roles/certificate/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22)
+are likely to be suitable for new contributors!
+
+**Code** is managed on [Github](https://github.com/linux-system-roles/certificate), using
+[Pull Requests](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests).
+
+### Python Code
+
+The Python code needs to be **compatible with the Python versions supported by
+the role platform**.
+
+For example, see [meta](https://github.com/linux-system-roles/certificate/blob/main/meta/main.yml)
+for the platforms supported by the role.
+
+If the role provides Ansible modules (code in `library/` or `module_utils/`) -
+these run on the *managed* node, and typically[1] use the default system python:
+
+* EL6 - python 2.6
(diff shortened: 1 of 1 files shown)


contributing.md
```diff
@@ -0,0 +1,31 @@
...
+  [Github](https://github.com/linux-system-roles/certificate), using [Pull
+Requests](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests).
+
+- The code needs to be **compatible with the Python versions supported by the role platform**.
+
+For example, see [meta](https://github.com/linux-system-roles/certificate/blob/main/meta/main.yml)
+for the platforms supported by the role.
+
+For example, EL6 requires python 2.6 support.  EL7 requires python 2.7 and python 3.6 support.  EL8 requires
+python 3.8 and later support.  EL9 requires python 3.9 and later support.
```
rjeffman, line 31: Isn't EL8 "official" Python version 3.6?

richm: Yes - but - Ansible generally prefers to use the python modularity releases on RHEL.  On RHEL 8.8/9.2, the version of Ansible is ansible-core 2.14 which uses the python311 module, even though RHEL 8 system python is 3.6, and RHEL 9 system python is 3.9.

rjeffman: I agree, but IPA, for example, is not supported with other than Python 3.6, and we currently rely on it for certificates.

I really don't know what to do in these cases. :-(

richm: > I agree, but IPA on EL8, for example, is not supported with other than Python 3.6, and we currently rely on it for certificates.
> 
> I really don't know what to do in these cases. :-(

You (as in the author of the certificate role) should have to do nothing in these cases.

Sorry - I should clarify - the version of python on the *control* node may change depending on the version of Ansible - this only affects things on the *control* side such as Ansible itself, or the use of plugins that run on the *control* side such as filters (e.g. `map`, `selectattr`, etc.) and tests (e.g. `is search`, `is regex`, etc.).  This usually does not affect the running of code on the *managed* node side, which almost always uses the system python.

rjeffman (approved):
LGTM.

contributing.md
```diff
@@ -0,0 +1,51 @@
...
+  Issues](https://github.com/linux-system-roles/certificate/issues). Issues labeled with
+[**help
+wanted**](https://github.com/linux-system-roles/certificate/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22)
+are likely to be suitable for new contributors!
+
+- **Code** is managed on
+  [Github](https://github.com/linux-system-roles/certificate), using [Pull
+Requests](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests).
+
+- The code needs to be **compatible with the Python versions supported by the role platform**.
```
richm, line 25: @rjeffman check the new text - I think it better explains the python support - let me know.

rjeffman: Text is much better now. Thank you for the change.

I hear, a lot, about the mess that Python versioning is right now, mostly due to legacy systems/older distros, and the current text make it a bit clearer.

rjeffman:
lgtm.

PR #177, opened 2023-06-22
test: easily generate certs for tests for other roles

richm:
This allows the certificate role to easily generate certs, keys
for use in tests of other system roles.  This eliminates the
anti-patterns like
https://github.com/linux-system-roles/logging/blob/main/tests/tests_server.yml#L16
where the test has to hard-code all of the paths that should be
only internal to the certificate role, and
https://github.com/linux-system-roles/logging/blob/main/tests/tests_server.yml#L212
where the role has to explicitly stop tracking the cert.

The caller of the role uses `certificate_test_mode: true` to activate this
feature.  I deliberately did not document this in the README because I don't
want end users to use this, only dev/qe.

The certs are returned in `certificate_test_certs`.  This is a `dict`.  Each
key is the cert name as passed in `certificate_requests`. This assumes `name`
is a name and not a path name, NSS DB key, etc.  The value is a `dict`.  The
dict has the following keys: `cert`, `cert_content`, `key`, `key_content`, `ca`,
`ca_content`.  The content keys are the `slurp`d b64decoded contents of the files,
the other keys are the file names.

The certs are not tracked after generation - the role uses `getcert stop-tracking`
on each generated cert.

If the caller sets `certificate_test_remove_files: true` then the role will
remove the files.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -133,3 +133,64 @@
   vars:
     __lsr_ansible_managed: "{{
       lookup('template', 'get_ansible_managed.j2') }}"
+
+# For using this role to issue certs for tests for other
+# system roles.  Assumes that the calling role wants
+# self signed certs e.g. as a replacement for various
+# openssl calls to generate certs, keys.  Also assumes
+# that each item in certificate_requests has a unique
+# name that can be used as the keys to the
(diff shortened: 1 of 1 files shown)


richm:
[citest]

PR #188, opened 2023-07-18
fix: Re-issue certificate if key size changes

rjeffman:
**Resolves:** 

When requesting a certificate key size is not evaluated to consider a new certificate has to be requested.

**Result:**

This patch adds 'key_size' to the metadata comparison to determine if a new certificate request must be performed.

**Issue Tracker Tickets (Jira or BZ if any):** [RHBZ#2186057](https://bugzilla.redhat.com/show_bug.cgi?id=2186057)

### DIFF
--- a/library/certificate_request.py
+++ b/library/certificate_request.py
@@ -67,9 +67,8 @@ options:
     required: false
   key_size:
     description:
-      - Generate keys with a specific keysize in bits.
+      - Generate keys with a specific keysize in bits, by default 2048.
     required: false
-    default: 2048
   ca:
     description:
       - CA that will issue the certificate. The available options
@@ -359,7 +358,7 @@ class CertificateRequestModule(AnsibleModule):
             directory=dict(type="str", default="/etc/pki/tls"),
             provider_config_directory=dict(type="str", default="/etc/certmonger"),
             provider=dict(type="str", default="certmonger"),
-            key_size=dict(type="int", default=2048),
+            key_size=dict(type="int", default=None),
             owner=dict(type="str"),
             group=dict(type="str"),
             mode=dict(type="raw"),
--- a/tests/tests_key_size_reissue.yml
+++ b/tests/tests_key_size_reissue.yml
@@ -1,0 +1,85 @@
+---
+- name: Test re-issue certificate if key size changes
+  hosts: all
+  become: false
+  gather_facts: false
+
+  tasks:
+    - name: Include role, ignore fail if certmonger version is not supported
+      block:
+        - name: Request certificate with key size
+          ansible.builtin.include_role:
+            name: linux-system-roles.certificate
+          vars:
+            certificate_requests:
+              - name: mycert_key_size
+                dns: www.example.com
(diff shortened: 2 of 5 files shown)


rjeffman:
[citest]

tests/tests_key_size_reissue.yml
```diff
@@ -0,0 +1,64 @@
...
+            subject_alt_name:
+              - name: DNS
+                value: www.example.com
+            key_size: 4096
+
+    - name: Issue certificate with key_size 3072
+      ansible.builtin.import_role:
+        name: linux-system-roles.certificate
+      register: result
+      failed_when: not result.changed
```
richm, line 40: I don't think this works - that is, `import_role` and `include_role` do not report `changed` if the last task reported `changed` - if you really want to test it this way, you'll need to register the result of `certificate_request` to a global role variable that you can export - we do something similar in the firewall role - https://github.com/linux-system-roles/firewall/blob/main/tasks/main.yml#L104 - then use it like https://github.com/linux-system-roles/firewall/blob/2e43f8a1e240d0b89775a61776d2377aa1c91113/tests/tests_zone.yml#L30


* add a `register: __certificate_request_result` to `name: Ensure certificate requests` - I think variables that are registered go into the public namespace, so you should be able to use it from the test
* in the tests, call the role with `include_role` (not `import_role`) and add `public: true` to export the privat

richm: hmm - well, the rhel 8.9 test passed - but I'm not even sure if the `failed_when` is being evaluated

rjeffman: yes, I forgot to remove the `register` and `failed_when` directives before opening the PR.

I'll fix it along with the CentOS/RHEL 7 fix.

rjeffman: fixed

rjeffman:
I fixed the issues with CentOS/RHEL 7, and rewrote the tests for `key_size` as they were not properly testing scenarios where certmonger version was less than `0.79.0`.

rjeffman:
[citest]

rjeffman:
[citest]

library/certificate_request.py
```diff
@@ -359,7 +358,7 @@ def _get_argument_spec():
             directory=dict(type="str", default="/etc/pki/tls"),
             provider_config_directory=dict(type="str", default="/etc/certmonger"),
             provider=dict(type="str", default="certmonger"),
-            key_size=dict(type="int", default=2048),
+            key_size=dict(type="int", default=None),
```
richm, line 361: so the certmonger provider will by default issue 2048 key_size?  On el7, el8, el9, and fedora?

rjeffman: Yes, by default it will be 2048.

Also, there's a warning in the role documentation that if the user wants to ensure a specific key size, the parameter `key_size` should be explicitly used as the default value could be changed depending on the role version.

richm (approved):

PR #190, opened 2023-08-09
ci: Add markdownlint, test_html_build, and build_docs workflows

spetrosi:
- markdownlint runs against README.md to avoid any issues with
  converting it to HTML
- test_converting_readme converts README.md > HTML and uploads this test
  artifact to ensure that conversion works fine
- build_docs converts README.md > HTML and pushes the result to the
  docs branch to publish dosc to GitHub pages site.
- Fix markdown issues in README.md

Signed-off-by: Sergei Petrosian <spetrosi@redhat.com>

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -38,6 +38,8 @@ jobs:
       - name: Convert role to collection format
         run: |
           set -euxo pipefail
+          # Remove to avoid running ansible-test on unrelated file
+          rm -f .pandoc_template.html5
           TOXENV=collection lsr_ci_runtox
           # copy the ignore files
           coll_dir=".tox/ansible_collections/$LSR_ROLE2COLL_NAMESPACE/$LSR_ROLE2COLL_NAME"
--- a/.github/workflows/markdownlint.yml
+++ b/.github/workflows/markdownlint.yml
@@ -1,0 +1,34 @@
+---
+# yamllint disable rule:line-length
+name: Markdown Lint
+on:  # yamllint disable-line rule:truthy
+  pull_request:
+  merge_group:
+    branches:
+      - main
+    types:
+      - checks_requested
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+permissions:
+  contents: read
+jobs:
+  markdownlint:
+    runs-on: ubuntu-latest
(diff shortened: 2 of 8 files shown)


.github/workflows/ansible-test.yml
```diff
@@ -38,6 +38,8 @@ jobs:
       - name: Convert role to collection format
         run: |
           set -euxo pipefail
+          # Remove to avoid running ansible-test on unrelated file
```
rjeffman, line 41: Can't this be fixed using [ignores](https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/ignores.html) ?

.github/workflows/markdownlint.yml
```diff
@@ -0,0 +1,34 @@
+---
+# yamllint disable rule:line-length
```
rjeffman, line 2: Is this needed?

(If I can I don't disable a linter rule.)

README.md
```diff
@@ -446,20 +427,14 @@ set to httpd.
...
 
 MIT
 
-
 ## Author Information
 
 Sergio Oliveira Campos (@seocam)
-
-
-[modeline]: # ( vim: set ts=2 sts=2 et: )
```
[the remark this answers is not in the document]
rjeffman: Agree!

.github/workflows/ansible-test.yml
spetrosi: .pandoc_template.html5 is a hidden file that is not present in the resulted collection that we publish to Ansible Galaxy and AH because we remove dotted files anyway, so adding an ignore for it is an overkill

.github/workflows/markdownlint.yml
spetrosi: I actually do not know, I copy-pasted this from other templates. @richm could you provide some info around this?

.github/workflows/ansible-test.yml
rjeffman: I'm fine with that.

.github/workflows/markdownlint.yml
rjeffman: If we are using this in all `.github` files (or non-ansible related files), than we should keep it for consistency.

PR #192, opened 2023-08-24
ci: Remove badges from README.md prior to converting to HTML

spetrosi:
- Remove thematic break after badges
- Remove badges from README.md prior to converting to HTML

Signed-off-by: Sergei Petrosian <spetrosi@redhat.com>


README.md
```diff
@@ -11,8 +9,6 @@ Linux system role to issue and renew SSL certificates.
 Basic usage:
 
 ```yaml
----
```
rjeffman, line 14: Why remove this part of the example? Isn't it better to have a full playbook?

If this is to be removed, maybe the `vars` part could be added at the role call, rather than a global setting.

spetrosi: This was a crazy `sed` overuse, fixed.

rjeffman (commented):
LGTM.

rjeffman (approved):

spetrosi:
Fixing the Ansible Plugin Scan check not seeing modules in tests/library in https://github.com/linux-system-roles/auto-maintenance/pull/295

PR #210, opened 2023-12-16
ci: Use supported ansible-lint action; run ansible-lint against the collection

richm:
The old ansible-community ansible-lint is deprecated.  There is a
new ansible-lint github action.  The new ansible-lint has several checks
related to ansible-test and the ignore files.  Many of our ignore settings
are not allowed any more and are required to be fixed or addressed in the
Ansible preferred way.

The python imports have to be wrapped in a try/except ImportError, and
where possible, an error must be returned from the module explaining
what was not able to be imported.

The result of this is that the .sanity files can be reduced to the
bare minimum which will greatly reduce the maintenance burden of
those files, make it easier to support newer versions of Ansible,
and make it easier to import the system roles collection into Galaxy
and Automation Hub.

`distutils.version` is deprecated, and it is a hard error in python 3.12
Instead, use `packaging.version` - unfortunately, this means having to
add a dependency on `python3-packaging` on platforms that support it -
notably, there is no python-packaging available on EL7, so fall back
to using `distutils.version` there.

The latest Ansible repo gating tests run ansible-lint against
the collection format instead of against individual roles.
We have to convert the role to collection format before running
ansible-test.

Role developers can run this locally using
`tox -e collection,ansible-lint-collection`
See https://github.com/linux-system-roles/tox-lsr/pull/125

Add `---` doc start to .markdownlint.yaml


richm:
[citest]

richm:
[citest]

richm:
centos 7 and 9 failures are baseos ci issues - the tests pass on those platforms locally

rjeffman (approved):
This is similar to what we did to ansible-freeipa a while ago.

LGTM.

library/certificate_request.py
```diff
@@ -108,62 +113,64 @@
...
+    type: str
   run_before:
     description:
       - Command that should run before saving the certificate.
-    required: false
+    type: str
   run_after:
     description:
       - Command that should run after saving the certificate.
-    required: false
```
spetrosi, line 166: Why are you removing `required: false`?

richm: The default is `required: false` - better style to omit `required` unless `required: true`

PR #243, opened 2024-11-26
fix: Workaround getcert issue when cert key-file is missing

rjeffman:
When trying to resubmit a certificate request that had the key file removed, "getcert" hangs indefinitely, causing the role to not respond.

This workaround checks if the private key file has been removed and fails ressubmiting the certificate request until the situation is fixed, by the user, using 'getcert stop-tracking --id <cert ID>'.

Issue Tracker Tickets (Jira or BZ if any): RHEL-13333

### DIFF
--- a/module_utils/certificate_lsr/providers/certmonger.py
+++ b/module_utils/certificate_lsr/providers/certmonger.py
@@ -40,6 +40,8 @@ else:
     HAS_DBUS = True
     DBUS_IMPORT_ERROR = None
 
+import os
+
 from ansible.module_utils.certificate_lsr.providers import base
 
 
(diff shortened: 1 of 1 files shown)


module_utils/certificate_lsr/providers/certmonger.py
```diff
@@ -254,6 +256,21 @@ def request_certificate(self):
         command = [getcert_bin]
 
         if self.exists_in_certmonger:
+            # if certificate exists in certmonger and key-file is missing,
+            # reissuing the certificate will hang certmonger.
+            # See: https://issues.redhat.com/browse/RHEL-69043
+            keyfile = self._certmonger_metadata.get("key-file")
+            if keyfile:
+                try:
+                    _fstat = os.stat(keyfile)
```
richm, line 265: Instead of `os.stat` with exception handling, you can do something like

```python
            if keyfile and not os.path.exists(keyfile):
                    self.module.fail_json(
                        "Resubmiting a request without the private key "
                        "file may hang certmonger. Please, stop monitoring "
                        "certificate '{0}' before reissuing.".format(
                            self._certmonger_metadata.get('nickname')
                        )
                    )
````

Comment (richm) [reply module_utils/certificate_lsr/providers/certmonger.py:265]: > I'm not sure I shouldn't process the stat result.
> 
> What if the file has been removed and a directory created?

ok, but in with the current code, `_fstat` is not used, and CodeQL flags it

Comment (rjeffman) [reply module_utils/certificate_lsr/providers/certmonger.py]: I change to use `os.path.isfile`. More than that and we would be looking into the file content, which I believe is way out of scope here.

File: module_utils/certificate_lsr/providers/certmonger.py [hunk 2]
```diff
@@ -254,6 +256,21 @@
...
```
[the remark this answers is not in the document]
rjeffman: Removed usage of `os.stat` and the extra variable.

richm:
[citest]

richm (approved):

richm:
looks like the Fedora failures are unrelated (problem with ipa?)

PR #281, opened 2025-08-13
test: reset hostname after ipa test

richm:
The tests_basic_ipa.yml test has to set a hostname for IPA
to work.  We need to reset the hostname after the test.

## Summary by Sourcery

Save the original hostname before setting the IPA test hostname and restore it after the tests on RedHat systems

Enhancements:
- Record the current hostname before changing it for the IPA test
- Add a teardown task to reset the hostname to its original value after the test


richm:
[citest]

spetrosi (approved):

PR #350, opened 2026-08-04
test: ignore Fedora and EL10 cert parsing error due to certmonger 309

richm:
Due to a bug fixed in https://codeberg.org/freeipa/certmonger/pulls/309
certmonger does not use the proper encoding for subject/common name
values which contain non-printable characters, such as used in
tests_subject_complex.yml. And the Fedora and EL10 pyca parser used by certtojson
is much stricter and does not like the encoding.  So ignore for now.

## Bug Fixes

* Improved certificate validation reliability for known encoding issues on Red Hat 10 and Fedora.
* Certificate parsing failures are now reported clearly with relevant error details.
* Certificate content checks are skipped when parsing is unavailable, preventing misleading validation results.
* Subject, SAN, key size, Key Usage, and Extended Key Usage checks now use successfully parsed certificate data.

### DIFF
--- a/tests/tasks/assert_certificate_parameters.yml
+++ b/tests/tasks/assert_certificate_parameters.yml
@@ -92,64 +92,89 @@
           {{ cert['owner'] | default('root') }}:
           {{ cert['group'] | default('root') }}
 
+    # NOTE: Due a bug in certmonger which is fixed by
+    # https://codeberg.org/freeipa/certmonger/pulls/309
+    # certmonger does not use the proper encoding for subject/common name
+    # values which contain non-printable characters, such as used in
+    # tests_subject_complex.yml. And the pyca parser used by certtojson
+    # in EL10 and Fedora is much stricter and does not like the encoding.
+    # So ignore for now.
     - name: Parse certificate
       certtojson:
         filename: "{{ cert['path'] }}"
       register: certificate_data
       changed_when: false
+      failed_when:
+        - certificate_data is failed
+        - not _is_affected or certificate_data.msg is not search(_msg)
+      vars:
+        _msg: >-
+          ParseError { kind: UnexpectedTag { actual: Tag { value: 13, constructed: true, class: Universal } } }
+        _is_affected: "{{ ((ansible_facts['distribution_major_version'] == '10' and
+          ansible_facts['os_family'] == 'RedHat'))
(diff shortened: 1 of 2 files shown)


richm:
[citest]

tests/tasks/assert_certificate_parameters.yml
```diff
@@ -92,64 +92,78 @@
...
-        fail_msg: expected value {{ expected }} != {{ actual }}
+        cert_issued: "{{ certificate_data.certificate }}"
+      block:
+        - name: Verify certificate subject
+          assert:
+            that:
+              - cert.subject | sort(attribute="name") ==
+                cert_issued.subject | sort(attribute="name")
+            fail_msg: >-
+              {{ cert.subject | sort(attribute="name") }} !=
```
spetrosi, line 136: What a user is expected to do in this case? To me this would seem like a weird error in the role.

richm: > What a user is expected to do in this case?

Not sure.  I guess using non-printable characters in the subject is uncommon.  This has been a problem in certmonger for a long time, and I guess the crypto libraries that read the cert/csr up until now have been relaxed in how they handle the encoding.  Using certs generated with the incorrect encoding will probably still work in most cases - so far it only seems to be a problem with the new strict python-cryptography parser in EL10.

> To me this would seem like a weird error in the role.

In certmonger, yes, not in the role.

richm:
[citest]

tests/tasks/assert_certificate_parameters.yml
spetrosi: Maybe print an error message with possible causes of this error - like certs created with incorerct encoding that you mentioned

richm:
[citest]

richm:
@spetrosi here is the message when cert parsing fails:

> Certificate parsing failed - this is likely due to a bug in certmonger which is fixed by https://codeberg.org/freeipa/certmonger/pulls/309 - message: error parsing asn1 value: ParseError { kind: UnexpectedTag { actual: Tag { value: 13, constructed: true, class: Universal } } }

spetrosi (approved):

richm:
rebased on top of fingerprint PR

PR #355, opened 2026-08-07
fix: ensure role is idempotent

richm:
Cause: The key_size from the request was being compared to the key_size read from the
existing certificate file in order to determine if they were different, even if the user
did not specify key_size in the request.

Consequence: If no key_size was specified in the request, the role would report `changed: true`
every time.

Fix: Do not use the `key_size` field for comparison if it was not specified in the request.

Result: The certificate role is idempotent.

Assisted-by: Cursor with models Cursor Grok 4.5, Composer 2.5, Sonnet 5.

This adds idempotency testing to the role.  A new role variable `certificate_is_changed`
is used internally to check if anything changed in the role.  Each role invocation in the
test is run twice to verify idempotency.  The tests_subject_complex has a known failure
on EL10 and later which will be fixed by https://codeberg.org/freeipa/certmonger/pulls/309

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved certificate comparison when optional key-size information is absent.
  - More accurately detects changes from certificate issuance, installation, trust updates, removals, downloads, and service-related operations.

- **Reliability**
  - Repeated role executions now correctly report no changes when the system is already up to date.
  - Trust-store additions and removals are tracked consistently.

- **Tests**
  - Expanded coverage for idempotency across certificate types, key sizes, trust configurations, hooks, and validation scenarios.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -2,6 +2,10 @@
 - name: Set version specific variables
   include_tasks: tasks/set_vars.yml
 
+- name: Set/reset certificate_is_changed
+  set_fact:
+    certificate_is_changed: false
+
 - name: Install and configure requirements and providers
   when: certificate_requests | length > 0
   vars:
@@ -14,6 +18,12 @@
         state: present
         use: "{{ (__certificate_is_ostree | d(false)) |
           ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
+      register: __certificate_packages_result
+
+    - name: Set certificate_is_changed
+      set_fact:
+        certificate_is_changed: true
+      when: __certificate_packages_result is changed  # noqa no-handler
 
     - name: Ensure provider packages are installed
       package:
@@ -25,6 +35,12 @@
       loop_control:
         loop_var: __certificate_provider
       when: __certificate_provider_vars[__certificate_provider].packages is defined
+      register: __certificate_provider_packages
+
+    - name: Set certificate_is_changed
+      set_fact:
+        certificate_is_changed: true
+      when: __certificate_provider_packages is changed  # noqa no-handler
 
     - name: Ensure pre-scripts hooks directory exists
       file:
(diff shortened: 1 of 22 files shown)


tests/tests_subject_complex.yml
```diff
@@ -10,12 +10,55 @@
...
+              is fixed by https://codeberg.org/freeipa/certmonger/pulls/309 -
+              message: {{ ansible_failed_result.msg }}
+          when: not _is_affected or ansible_failed_result.results |
+            selectattr('msg', 'search', _msg | regex_escape) | list | length == 0
+          vars:
+            _msg: >-
+              ParseError { kind: InvalidValue, location: [1, 0, "AttributeTypeValue::value", "AttributeValue::PrintableString"] }
+            _is_affected: "{{ ((ansible_facts['distribution_major_version'] == '10' and
+              ansible_facts['os_family'] == 'RedHat'))
+              or ansible_facts['distribution'] == 'Fedora' }}"
```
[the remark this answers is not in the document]
richm: I want the test to fail - I want whoever runs this on EL11 to have to stop and figure out why it is still failing.  In fact, I'm hoping the certmonger fix lands soon so I can remove this hack.

tasks/main.yml
```diff
@@ -14,6 +18,12 @@
         state: present
         use: "{{ (__certificate_is_ostree | d(false)) |
           ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
+      register: __certificate_packages_result
+
+    - name: Set certificate_is_changed
+      set_fact:
+        certificate_is_changed: true
```
spetrosi, line 25: `certificate_is_changed: true` is set for test purposes only?

richm: > `certificate_is_changed: true` is set for test purposes only?

I'm using it for test purposes, but I suppose it could be used for other purposes.

There really is no other way to test for idempotency without checking all of the changed tasks in a role :-(

richm:
[citest]

richm:
[citest]

PR #359, opened 2026-08-13
feat: Add argument spec validation to Certificate role

DonatSzabo:
Enhancement: Added argument spec validation to certificate role. Also wrote tests for  it found in tests/tests_invalid_input.

Reason: Because it is a good addition to the linux-system-roles project.

Result: Successfully added it and prepared tests for it. I used AI during this implementation.

Issue Tracker Tickets (Jira or BZ if any): https://github.com/linux-system-roles/postfix/issues/206 https://redhat.atlassian.net/browse/RHELMISC-16008

Notes: Some values (eg.: dns, email...) get passed as raw, but these are catched later on. Please review and advise!

## Summary by CodeRabbit

* **New Features**
  * Added validation for certificate issuance, renewal, waiting, and trust-store configuration.
  * Added validation for certificate metadata, SANs, ownership, permissions, hooks, providers, and trust sources.
  * Added clear rules for required fields, supported states, data types, and trust-source configuration.
  * Added cross-field validation for certificate request options.

* **Documentation**
  * Documented validation through argument specifications and role tasks.

* **Tests**
  * Added coverage for invalid, incomplete, and incorrectly typed certificate configuration.


DonatSzabo:
I'm working on fixing the bugs and changing tests_invalid_input.

meta/argument_specs.yml
```diff
@@ -0,0 +1,152 @@
...
+            type: str
+            description: >-
+              Organizational unit requested for the certificate subject.
+          contact_email:
+            type: str
+            description: >-
+              Contact email requested for the certificate subject.
+          key_usage:
+            type: list
+            elements: str
```
richm, line 100: The module has this, which seems to imply that you can specify that each element must be one of the list of choices:

```python
            key_usage=dict(
                type="list",
                choices=KEY_USAGE_CHOICES,
                default=KEY_USAGE_DEFAULTS,
                elements="str",
            ),
```

Looking at the Ansible role argument spec docs, it is unclear if this works.  Can you do some investigation to find out if you can specify choices in the role argument_specs.yml, and if so, does it really enforce that each list element must be one of the choices?

richm:
In the auditd role https://github.com/linux-system-roles/auditd/blob/main/tasks/assert_role_vars.yml is used to validate the role arguments which are of type `raw`, or require more complex checking than just "is this thing an int?" or where you can't easily specify the choices. I think you should add an assert_role_vars.yml.  This is also useful in case someone is using ansible 2.9 where meta/argument_specs.yml is not supported.

tests/tests_invalid_input.yml
```diff
@@ -0,0 +1,228 @@
...
+            that:
+              - __valid_defaults_succeeded | default(false)
+            fail_msg: >-
+              Role should succeed with default parameters
+
+        # --- certificate_requests validation ---
+        # Role argument validation requires Ansible 2.10+; skipped on older versions.
+
+        - name: Argument specs reject missing name in certificate_requests
+          when: ansible_version.full is version('2.10', '>=')
```
richm, line 28: instead of having multiple tasks with this same `when` condition, you should refactor this into another `block` with this `when` condition, and put all of the other `block`/`rescue` under that outer `block`.

meta/argument_specs.yml
DonatSzabo: I have added the choices to argument specs. I have checked and I am pretty sure that it goes and iterates trough every list item. I also added a test for key usage  choice and they work correctly. The test passes 1 valid and 1 invalid input, the valid one gets passed but the invalid one doesn't

tests/tests_invalid_input.yml
DonatSzabo: Done

spetrosi:
[citest]

spetrosi (approved):

tests/tests_invalid_input.yml
```diff
@@ -0,0 +1,334 @@
...
+              ansible.builtin.assert:
+                that:
+                  - __invalid_input_trust_state_failed | default(false)
+                fail_msg: >-
+                  meta/argument_specs should reject certificate_trust state
+                  values outside the allowed set (present, absent)
+
+        # --- certificate_trust cross-field validation (assert in tasks, not argument_specs) ---
+
+        - name: Assert rejects trust entry with no source when state is present
```
richm, line 228: Where is this checked?  I don't see this in assert_role_vars.yml or tasks/main.yml

DonatSzabo: Its in tasks/certificate_trust.yml

name: Validate certificate_trust items
  assert:
    that:
      - item.name is defined
      - item.state | d('present') in ['present', 'absent']
      - >-
        item.state | d('present') == 'absent' or
        [item.content is defined, item.src is defined, item.url is defined] |
        select | list | length == 1
    fail_msg: >-
      Each certificate_trust item must have a name, state must be 'present'
      or 'absent', and exactly one of content, src, or url must be given
      when state is 'present'
  loop: "{{ certificate_trust }}"
  loop_control:
    label: "{{ item.name | d('unnamed') }}"
  register: __certificate_trust_validated

DonatSzabo: Should I add this to assert_role_vars?

richm: > Its in tasks/certificate_trust.yml

ah, ok  - it was already there, not in your PR

> Should I add this to assert_role_vars?

No, it is fine as it is

richm:
You'll need to rebase on top of the latest `main` branch and resolve any conflicts.  Then I think it will be ok to merge.

spetrosi:
[citest]

spetrosi:
RHEL 7 tests are failing. Gemini told that ansible 2.9 has older jinja which didn't have is integer. This is correct because indeed some syntax in Ansible is passed from Jinja.
AI suggests using `my_var is number` or `(my_var | type_debug) == 'int'` , please investigate which is better.

tasks/assert_role_vars.yml
```diff
@@ -69,9 +69,7 @@
     that:
       - >-
         item.mode is string
-        or (item.mode is integer
-            and item.mode is not sameas true
-            and item.mode is not sameas false)
```
spetrosi: It's weird that item.mode is and item.mode == is inconsistent. Let's use `==` and check for both in a single go:
item.mode in ['string', 'int']

^^ that's a pseudo code, Ansible might define 'string' as 'str'.

DonatSzabo: Okay, makes sense. I will get on that!

spetrosi:
[citest]

tasks/assert_role_vars.yml
richm: you can test this locally using our el7 ansible 2.9 container image:

```bash
> podman run --rm -it --entrypoint /bin/bash quay.io/linux-system-roles/lsr_ansible_el7:latest
[root@a392f466f0a1 /]# ansible -c local -i localhost, -e myvar=myvalue all -m debug -a msg='{{myvar|type_debug}}'
localhost | SUCCESS => {
    "msg": "unicode"
}
[root@a392f466f0a1 /]# ansible -c local -i localhost, -e myvar=myvalue all -m debug -a msg='{{myvar is string}}'
localhost | SUCCESS => {
    "msg": true
}
[root@a392f466f0a1 /]# ansible -c local -i localhost, -e myvar=myvalue all -m debug -a msg='{{myvar is number}}'
localhost | SUCCESS => {
    "msg": false
}
```

richm: note that by default `-e var=value` will pass in `var` as a `string` valued variable - if you want to test with a real integer, pass in the JSON style definition:

```bash
[root@a392f466f0a1 /]# ansible -c local -i localhost, -e '{"myvar": 42}' all -m debug -a msg='{{myvar is number}}'
localhost | SUCCESS => {
    "msg": true
}
```

DonatSzabo:
[citest]

tasks/assert_role_vars.yml
```diff
@@ -67,9 +67,7 @@
 - name: Assert mode is a string or integer
   ansible.builtin.assert:
     that:
-      - >-
-        item.mode is string
-        or (item.mode | type_debug) == 'int'
+      - (item.mode | type_debug) in ['str', 'int']
```
richm: `      - (item.mode | type_debug) in ['str', 'int', 'unicode']`

This should make it work on el7

richm:
[citest]

richm:
[citest]

richm (approved):

richm:
@spetrosi ok to merge?

richm:
[citest]

tests/tests_invalid_input.yml
```diff
@@ -0,0 +1,400 @@
...
+            that:
+              - __valid_defaults_succeeded | default(false)
+            fail_msg: >-
+              Role should succeed with default parameters
+
+        # --- argument_specs validation ---
+        # Role argument validation requires Ansible 2.10+; skipped on older versions.
+
+        - name: Run argument specs validation tests
+          when: ansible_version.full is version('2.10', '>=')
```
richm:
```suggestion
          when: ansible_version.full is version('2.11', '>=')
```

meta/argument_specs.yml
```diff
@@ -0,0 +1,190 @@
...
+            description: >
+              The name of the certificate. A full path can be used to
+              choose the directory where files will be stored.
+          ca:
+            type: str
+            required: true
+            description: >
+              The CA that will issue the certificate (e.g. self-sign, ipa).
+          dns:
+            type: raw
```
richm, line 35: why are `dns`, `email`, and `ip` using type `raw`?  Is it because they can be `null`?

richm (changes requested):

meta/argument_specs.yml
DonatSzabo: Its because these fields can accept a string or a list of strings.

richm (approved):

richm (approved):

